PR #78, opened 2012-11-27
Auto detect version of grails/griffon when within a grails/griffon app

sjurgemeyer:
This pull request adds scripts for grails and griffon that can auto detect the version of an application based on application.properties.  The feature is off by default, but can be turned no by adding GVM_AUTODETECT=1 in .bashrc/.zshrc.

When the option is turned on 'grails', 'grails-debug','griffon' and 'griffon-debug' will first check application.properties for the proper version, then call 'gvm use' for that version before calling grails/griffon.

There are also a grails-nodetect and griffon-nodetect scripts included to just use whatever version of grails or griffon is current for GVM even if you're in an application.  This will be useful for things like 'grails upgrade'.

Let me know if there are any items that are unclear or need to be cleaned up.  I hope this is helpful.


snimavat:
@marc0der was this ever done for sdkman ?

marc0der:
Yes, it is available as an extension in the sdkman-extensions project. I feel that it should never become a part of the core as the core will always remain candidate agnostic.

PR #189, opened 2013-07-15
A putative solution to Issue 187.

russel:
This small patch avoids writing new environment variables in the case that the user already has the "HOME" variable already defined. It is assumed that if they have the "HOME" variable defined then they have appropriately set their PATH variable.


marc0der:
Cleaning up, can I close this one?

marc0der:
No response so closing.

PR #211, opened 2013-08-23
Use zsh-style word splitting when running on zsh; remove zsh-specific setopt

marshallpierce:


rgoldberg:
This PR seems obsolete.

PR #293, opened 2014-09-15
Check for CANDIDATE_DIR exist

haomingz:
if `$CANDIATE_DIR` exist, then add to `PATH`. 
Reduce PATH and other env variables size.


marc0der:
This was fixed in a subsequent refactor. Candidate homes are only added to the path if the candidate is installed.

PR #300, opened 2014-10-20
add "export" on GVM_CANDIDATE_COUNT

aks:
The `gvm current` command was failing (in a bash environment) with GVM_CANDIDATE_COUNT having no value.

The insertion of an "export" on the assignment of this variable fixes the problem.

### DIFF
--- a/src/main/bash/gvm-init.sh
+++ b/src/main/bash/gvm-init.sh
@@ -165,7 +165,7 @@ fi
 # In bash arrays are zero index based, in zsh they are 1 based(!)
 gvm_set_candidates
 if [[ -z "$ZSH_VERSION" ]]; then
-	GVM_CANDIDATE_COUNT=${#GVM_CANDIDATES[@]}
+	export GVM_CANDIDATE_COUNT=${#GVM_CANDIDATES[@]}
 else
 	GVM_CANDIDATE_COUNT=${#GVM_CANDIDATES}
 fi


nobeans:
#### GVM_CANDIDATE_COUNT doesn't need

I've also confused such a behavior of the variables. But now it's solved for me.

`GVM_CANDIDATE_COUNT` is so unnecessary that it can be removed. Currently, it's used to wrap a difference between zsh and bash, like:

* zsh: `${#array[@]}`
* bash: `${#array}`

But, in fact, the above bash style syntax also works for zsh.

So, you can simply use `${#GVM_CANDIDATES}` instead of `${GVM_CANDIDATE_COUNT}`. It's a very simple solution for me. Actually, some scripts in gvm-cli are doing so.

#### Why is GVM_CANDIDATE_COUNT empty?

Both `GVM_CANDIDATES` and `GVM_CANDIDATE_COUNT` can be used in a top-level shell like a login shell.
But `GVM_CANDIDATE_COUNT` cannot be used in a sub shell (including a tmux shell), because when an initialization is already done, gvm-init.sh would return without declaring of `GVM_CANDIDATE_COUNT` at https://github.com/gvmtool/gvm-cli/blob/master/src/main/bash/gvm-init.sh#L129.

#### How to reproduce

In short, gvm-current.sh doesn't work in a sub-shell. You can easily reproduce it as follows:

```
login-shell$ gvm current
(works well)
login-shell$ zsh
sub-shell-zsh$ gvm current
__gvmtool_current:11: bad math expression: operand expected at end of string
sub-shell-zsh$ exit
login-shell$ bash
sub-shell-bash$ gvm current
bash: ((: i <= : syntax error: operand expected (error token is " ")
No candidates are in use
```

#### How to fix (I recommend)

It can be fixed to replace `${GVM_CANDIDATE_COUNT}` with `${#GVM_CANDIDATES}`. I've also prepared to send a PR about it to gvm-cli ;-)

Besides, you should also run a test in a sub-shell.

marcoVermeulen:
@nobeans sounds great! if you raise me a pull request, I'll review and merge it.

marcoVermeulen:
And then I saw that you already had done so :-) Thanks!

marcoVermeulen:
@nobeans I've looked again and don't see a pull request for this. The solution sounds sensible and will remove the age-old hack.

Regarding the tests, it would be nice, but we don't really have the means to do this. We are using a `java.lang.Process` to run all the tests. As far as I know it doesn't have the notion of sub-shells.

nobeans:
Sent PR: https://github.com/gvmtool/gvm-cli/pull/316

Hmm, it seems difficult to test of sub-shells. Specking of testing shell, I know https://github.com/kiy0taka/spock-shell. But I don't know if it is fit for the purpose.

marcoVermeulen:
Don't worry about testing it, as long as the standard suite still passes
all is well. I'll also do some manual testing before I release.

marcoVermeulen:
Okay, looked it over and it looks good. I'll release it shortly.

I recall why we had to do this hack: It was because earlier versions of zsh were inconsistent with bash, necessitating us to implement this workaround.

nobeans:
> I recall why we had to do this hack: It was because earlier versions of zsh were inconsistent with bash, necessitating us to implement this workaround.

Oh, it sounds tough...

BTW, I think now you can just close this PR.

PR #302, opened 2014-10-29
Fix creating symlinks in cygwin

busches:
mklink function from: http://stackoverflow.com/questions/3343988/accessing-a-cygwin-symlink-from-windows/11166347#11166347


busches:
Tried the latest Cygwin and Sdkman on Windows 10 to see if this was still an issue and it is.  I've rebased it so it can be merged in. Let me know if there's anything else I need to do to get this merged, seems to be impacting a few people from the comments in the issues. Thanks!

PR #306, opened 2014-11-19
Remote version is empty.

abcfy2:
Fix this issue:

```
The current version is , but you have 2.2.0.
```

### DIFF
--- a/src/main/bash/gvm-broadcast.sh
+++ b/src/main/bash/gvm-broadcast.sh
@@ -17,17 +17,13 @@
 #
 
 function __gvmtool_broadcast {
-	if [ "${BROADCAST_HIST}" ]; then
-		echo "${BROADCAST_HIST}"
-	else
-        if [[ "$GVM_FORCE_OFFLINE" == "true" ]]; then
-            BROADCAST_LIVE=""
-        else
-            BROADCAST_LIVE=$(curl -s --connect-timeout 1 "${GVM_BROADCAST_SERVICE}/broadcast/latest")
-            gvm_check_offline "$BROADCAST_LIVE"
-        fi
-        echo "${BROADCAST_LIVE}"
-	fi
+    if [[ "$GVM_FORCE_OFFLINE" == "true" ]]; then
+        BROADCAST_LIVE=""
+    else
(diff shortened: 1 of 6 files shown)


marcoVermeulen:
Sorry for not replying earlier. This changes the behaviour of GVM significantly. Since this radical change of behaviour was never discussed upfront (and also I don't agree with), I cannot merge the PR.

GVM functions much like a browser does, only in the shell. It makes regular calls back home for several reasons:

* checks if new broadcasts are available, a core part of gvm
* makes sure that you are still online, otherwise goes into offline mode
* checks daily to see if new updates are available

I think that removing these checks will break gvm in many other places since it is vital for it to remain connected to the internet. Once again, would you try using a browser while being offline? Well neither would you do so with GVM as much of it's logic lives on the serverside. It's not merely a script application, it is a client-server application.

Hope the explanation makes sense!

abcfy2:
OK I know. I make these changes because connect to ``api.gvmtool.net`` is too slow from here, and also have some bugs(remote version is empty and update failed because download not complete, but the available gvm has been removed), often timeout or hang. I have to spend much time at every time to use gvm. So I want gvm will use offline resources as much as possible. It's unnecessary for gvm to connect internet at every time. The browser also holds local caches for speed up, isn't it?

But you are right. My changes will cause gvm never check update unless ``gvm selfupdate``. Maybe check update and broadcast in ``gvm-init.sh`` at every day is better.

Regards.

marcoVermeulen:
Many of the issues you were experiencing have been fixed now. They are on
DEV and I'll be rolling them out shortly.

I have fixed remote version being blank (because of multiple shell issue).
Also, broadcasts are now very lightweight calls that only retrieve a
broadcast id, no longer the entire broadcast message.

Unfortunately we do need to ensure that you are online because most
commands in GVM rely on the internet. Here I refer to commands like list,
install, default, broadcast etc. which all rely on dynamic content from the
API. Also, browsers only cache static content.

abcfy2:
Ah ha, it's right. I've already considered(Only not checking update and broadcast automatic).

    Usage: gvm <command> <candidate> [version]

``version`` is optional, if not provide, checking online. Otherwise check local if exits, find in local, use it, if not, connect the internet and download. This is my basic idea.

The candidate with specified version is static content. :)

PR #315, opened 2015-02-06
avoid repeated execution when start a new shell (in tmux or screen)

wonder365:


wonder365:
in my cygwin, gvm-init.sh use about 3 seconds. when I use tmux create multi terminal, it blocks each time

marc0der:
This was addressed some time ago. We only ever source the module scripts once per shell. I can confirm that this is now working on tmux.

PR #318, opened 2015-03-08
Add 'outdated' command.

nobeans:


marcoVermeulen:
Thanks, looks great. The build will propagate it to the dev environment, after which I'll give it a quick test before taking it live tomorrow or later today.

marcoVermeulen:
I've deployed on dev, and found one or two small things that could be addressed to improve your PR.

It would be nice to have a small label `Outdated:` included in the output, to make it consistent with the `current` command. Something like this:

    % gvm outdated
    Outdated:
    gradle (2.2.1 < 2.3)
    groovy (2.4.0, 2.3.9 < 2.4.1)

Also, if it is found that all candidates are up-to-date, it would be nice to see console output like:

    % gvm outdated
    All candidates are up-to-date

Lastly, we need to update the `help` command to reflect the new `outdated` command. We also need to think of giving it an alias like `o`, similar to all the others. And cukes where it makes sense.

Apart from these little things, you have done an awesome job. Thanks so far!

nobeans:
Sent the PR having two commits: the one is about the label in output and another one is a help and a shortcut with the cukes.

marcoVermeulen:
Great, thank you! Everything looks perfect. I'll be releasing it soon.

PR #321, opened 2015-03-10
Fix the issue that the "Outdated:" label is not displayed in some cases.

nobeans:
Sorry, but the current outdated command can't print the "Outdated" label properly in some case: In order, the first candidate is not outdated, but the after-second candidate is outdated.

It's caused by the wrong condition.

### DIFF
--- a/src/main/bash/gvm-outdated.sh
+++ b/src/main/bash/gvm-outdated.sh
@@ -46,7 +46,7 @@ function __gvmtool_outdated {
                 ;;
             *)
                 if [ -n "${outdated}" ]; then
-                    [ ${installed_count} -eq 0 ] && echo "Outdated:"
+                    [ ${outdated_count} -eq 0 ] && echo "Outdated:"
                     echo "${outdated}"
                     (( outdated_count += 1 ))
                 fi


nobeans:
```
GVM_CANDIDATES=asciidoctorj crash gaiden glide gradle grails griffon groovy groovyserv jbake lazybones springboot vertx
```

* The first installed candiate in the above order is outdated, this issue isn't reproduced.
* The first installed candiate in the above order is not outdated, this issue is reproduced.

I could reproduce this issue by the following test.

```
  Scenario: PR#321
    Given  the candidate "grails" version "1.3.9" is already installed and default
    And the default "grails" candidate is "2.4.4"
    And the candidate "groovy" version "2.0.5" is already installed and default
    And the default "groovy" candidate is "2.0.5"
    When I enter "gvm outdated"
    Then I see "Outdated:"
    And I see "grails (1.3.9 < 2.4.4)"
    And I do not see "groovy"
```

And it passed after the branch of this PR.

nobeans:
Hmm, I've just realized that my explanation and the test code is inconsistent.

I though the order of GVM_CANDIDATES is `asciidoctorj crash gaiden glide gradle grails griffon groovy groovyserv jbake lazybones springboot vertx` and it seems true in production, at least in my computer. But in the test, the order is `groovy grails` and so the issue is reproduced.

nobeans:
Should I add the test case?

marcoVermeulen:
Let's leave the test case then. It seems more trouble than it's worth. I'll merge it for you as it is.

Issue #256, opened 2013-12-17
Listings are not in numerical order


russel:
|> gvm list gradle
# 
# Available Gradle Versions
- 1.9                  0.7  
   1.8  
   1.7  
   1.6  
   1.5  
   1.4  
   1.3  
   1.2  
  > \* 1.10  
     1.1  
     1.0  
     0.9.2  
     0.9.1  
     0.9  
     0.8  
# 
- \- local version
- \- installed
  > - currently in use
  ================================================================================

Issue #104, opened 2012-12-31
Semantic sort order for versions


glaforge:
A very minor, non prioritary feature request: use a semantic sort ordering for versions.
If you use the usual lexicographical order, 1.7.10 will appear after 1.7.1 and before 1.7.2, but the real order of releases isn't so.

Example, when you list Groovy versions:

```
...
1.7.9
1.7.8
1.7.7
1.7.6
1.7.5
1.7.4
1.7.3
1.7.2
1.7.11
1.7.10
1.7.1
1.7.0
...
```

So the idea would be to have a special comparator that would compare versions properly with separating the version elements.

Issue #73, opened 2012-11-20
Ordering of list command versions not consistent when exceeding 9


marc0der:
Example:
 * 1.7.0
 * 1.7.1
 * 1.7.10
 * 1.7.11
   1.7.2
   1.7.3

PR #323, opened 2015-03-24
Adding semantic sorting of 'gvm list [candidate]' versions

mbjarland:
This PR sorts the versions in 'gvm list [candicate]' semantically (numerically) as you would expect. The code should also handle non-parseable version tokens and different lengths for versions, eg ["2.3.15", "4", "3.5-SNAPSHOT"]. 

This PR should address issues #73 #104 #256 #268 which all seem to be different variations of this problem. 

The existing code does alpha sorting which gives you [3.2.1, 3.2.11, 3.2.2], i.e. sorts 11 in between 1 and 2. 

This PR contains the following changes (all to server.groovy): 
- semantic sorting of versions
- rewrite of the prepareListView method. The new method is a tad more functional, faster (about 3-5x), and it flows the versions correctly across the given number of columns. 
- removed the hard coded number of rows (COLUMN_LENGTH) and replaced it with NUMBER_OF_COLUMNS as I think specifying the number of columns feels more natural. 
- created a cucumber test to validate ordering, but need to figure out why it is not being run. Any pointers much appreciated.

A number of the tests (for things other than my change) are breaking when I execute them. Not sure if this is a local problem on my machine or just the state of affairs. 

Also initially failed to start the server locally because of the following exception: 

```
java.lang.IllegalArgumentException: Invalid module identifier: vertx.mongo-persistor-v1.2. Should be of form owner~name~version
...
```

so I changed the mongo-persistor dependency to say: 

``` groovy
//server.groovy
container.deployModule 'io.vertx~mod-mongo-persistor~2.1.0', config
```

which gets us past the server startup issue. 

After adding the current groovy versions and some bogus test values to try to break things into my mongo db and running the server, we get the below output after hitting: 

http://localhost:8080/candidates/groovy/list?platform=Darwin&current=2.3.8&installed=2.3.8

``` none
Available Groovy Versions
     5.a.11               2.3.3                2.0.8                1.8.1          
     3.0.1                2.3.2                2.0.7                1.8.0          
     3.0-SNAPSHOT         2.3.1                2.0.6                1.7.11         
     3                    2.3.0                2.0.5                1.7.10         
     2.90                 2.2.2                2.0.3                1.7.9          
     2.4.3                2.2.1                2.0.2                1.7.8          
     2.4.2                2.2.0                2.0.1                1.7.7          
     2.4.1                2.1.9                2.0.0                1.7.6          
     2.4.0                2.1.8                2.b.8                1.7.5          
     2.3.11               2.1.7                1.8.9                1.7.4          
     2.3.10               2.1.6                1.8.8                1.7.3          
     2.3.9                2.1.5                1.8.7                1.7.2          
 > * 2.3.8                2.1.4                1.8.6                1.7.1          
     2.3.7                2.1.3                1.8.5                1.7.0          
     2.3.6                2.1.2                1.8.4                               
     2.3.5                2.1.1                1.8.3                               
     2.3.4                2.1.0                1.8.2                               
+ - local version
* - installed
> - currently in use
```


marcoVermeulen:
Thanks for the great work. I did raise one or two concerns, namely the reformatting of `server.groovy` as well as the cucumber spec. Apart from that it looks good. If you fix those things, I will merge it.

Just to make you aware that even if your change makes it in, all this code will be completely dropped within the next month or two :-(

This is due to the fact that the entire API will be re-written with spring boot or play as opposed to vertx. I also plan to introduce a domain model that will automatically handle semantic sorting of the list using a comparator.

mbjarland:
Formatting fixed by reverting it back to the original form. I am not all groovy with cucumber (I am to use spock), so fixing the spec might take me a few days. 

No worries about dropping the code, this was a fun exercise.

mbjarland:
So I took a look at the cucumber tests and modified to use a data table to specify the expected table of values. However, running gradlew gives the following test output: 

```
867 tests completed, 22 failed, 15 skipped
```

which in addition to having failing tests does not contain my test. As mentioned I am a tad unfamiliar with cucumber, but I noticed there is a `@manual` tag on the `list_candidates.feature`. I assume this means it is not included in the normal test run from gradle. Could you give me a pointer as to what the philosophy here is and how I would go about running my cucumber test?

marcoVermeulen:
Right you are! That @manual must have crept in there at some earlier stage.
I wouldn't suggest using a datatable, but keeping it as simple as possible.
I would go for:

*Then I see:*
*"""*
*================================================================================*
*Available Gradle Versions*
*================================================================================*
* > * 2.3                  1.10
        *
*     2.2.1                1.1
         *
*     2.2                  1.0
         *
*     2.1                  0.9.2
         *
*     2.0                  0.9.1
         *
*     1.9                  0.9
         *
*     1.8                  0.8
         *
*     1.7                  0.7
         *
*     1.6
        *
*     1.5
        *
*     1.4
        *
*     1.3
        *
*     1.2
        *
*     1.12
         *
*     1.11
         *

*================================================================================*
*+ - local version*
** - installed*
*> - currently in use*
*================================================================================*
*""" *

In other words, do the setup adding all the versions one by one that you
want to make part of your test, then simply assert that the output matches
what you expect to see.

The step def for *Then I see* is already defined in
*command_line_interop_steps.groovy* as:
*And(~'^I see \"([^\"]*)\"$') { String output ->*
*    assert result.contains(output)*
*}*

Hope this helps some.

marcoVermeulen:
Right you are! That @manual must have crept in there at some earlier stage.
I wouldn't suggest using a datatable, but keeping it as simple as possible.
I would go for:

    Then I see:
    """
     > * 2.3                  1.10
         2.2.1                1.1
         2.2                  1.0     
    """

In other words, do the setup adding all the versions one by one that you
want to make part of your test, then simply assert that the output matches
what you expect to see.

The step def for `Then I see` is already defined in `command_line_interop_steps.groovy` as:

    And(~'^I see \"([^\"]*)\"$') { String output ->
        assert result.contains(output)
    }

Hope this helps some.

mbjarland:
Might be my unfamiliarity with cucumber, but I am unable to get the test to work. And I do not mean that my code breaks but rather that the cucumber tests fail to execute or fail for causes I do not have the time or insight to figure out. 

At this point I think this PR is either dead in the water for the time being or needing some attention from somebody more familiar with the setup. Removing the @manual tag causes the test to fail with HTTP errors from jetty etc.

About the feature format: Specifying a verbatim string for the returned result is more brittle than a data table as it will break if there is so much as a single space difference in padding etc. Not sure I agree with the verbatim format being better or simpler. Then again, perhaps I'm missing something so feel free to enlighten me if that is the case. 

Would have been nice to get this in. I have tried to accommodate project standards, reverting the formatting to preferred indentation, attempted writing a cucumber spec, ran the server locally to see that the sorting works etc etc, but there is a limit on the amount of time you would realistically expect a contributor to spend on figuring out scaffolding things. I think I just reached mine for now. 

If anybody feels like resurrecting this, please do so. Otherwise I will leave this PR as is.

src/test/cucumber/gvm/list_candidates.feature
```diff
@@ -37,3 +37,25 @@ Feature: List Candidates
     When I enter "gvm list groovy"
     Then I see "Available Groovy Versions"
     And I see " > + 2.2-SNAPSHOT"
+
+  Scenario: List all abailable versions and validate sorting
```
grimreaper: nit: available

marc0der:
Closing this as it has long since been implemented on the serverside.

PR #325, opened 2015-03-29
Use ${GVM_DIR} instead of hardcoded .gvm

LeeWorrall:
__gvmtool_determine_current_version currently has ".gvm" hardcoded instead of using the GVM_DIR env var, this pull request fixes that.


marcoVermeulen:
Good spot, I'll probably release soon.

PR #336, opened 2015-07-03
Do not modify Java environment in Cygwin. Fixes #335

### LINKED ISSUE #335: Cygwin: gvim-init.sh breaks Java environment variables
Cygwin provides a Unix environment usable for utilities built against the Posix layer of Cygwin. However, I am not aware that there is any JDK available, that has been ported to Cygwin (it might be possible to do so with OpenJDK). So, why does gvm-init.sh modify the Java environment variables (e.g. JAVA_HOME and CP) turning them into Unix syntax, if every available java.exec is a plain Windows application that does not understand Unix paths?

Please leave these variables alone. They can break Java applications called from the Cygwin's bash.

joehni:


rgoldberg:
This is obsolete.  The changes herein were committed in d700b2175e8ed6ba222af4dafd1ee37d8583e1a9 & merged into `master`.

PR #342, opened 2015-07-27
Added a flag "GVM_NOT_REFRESH" in src/main/bash/gvm-init.sh to avoid curl hang on init when there is no network

liuchong:
The flag will tell gvm not to refresh to check if is up to date, because it will very slow if network is slow or there is no network, the curl will keep waiting for response.


liuchong:
so, if I do not want curl access the remote urls, I will make a "export GVM_NOT_REFRESH=true"(or ="yes"...) in my .bashrc before "source path/to/.gvm/bin/gvm-init.sh"

PR #343, opened 2015-07-30
Fix init from missing the last few candidates

danthegoodman:
In bash, `${#ARR}` returns the length of the string in the first slot.
'asciidoctorj' has a length of 12, which was working until more
candidates were added.

`${#ARR[*]}` is the correct way to return the length of an array in
bash. Thankfully, this works in ZSH as well.

This issue is preventing `vertx` from appearing on my path.

To demonstrate, you may run these commands in your shell:

```
$ echo ${GVM_CANDIDATES[*]}
asciidoctorj crash gaiden glide gradle grails griffon groovy groovyserv 
jbake jbossforge lazybones springboot vertx

$ echo ${#GVM_CANDIDATES}
12

$ echo ${#GVM_CANDIDATES[*]}
14
```

### DIFF
--- a/src/main/bash/gvm-init.sh
+++ b/src/main/bash/gvm-init.sh
@@ -170,7 +170,7 @@ if [[ "${GVM_INIT}" != "true" ]]; then
     # The candidates are assigned to an array for zsh compliance, a list of words is not iterable
     # Arrays are the only way, but unfortunately zsh arrays are not backward compatible with bash
     # In bash arrays are zero index based, in zsh they are 1 based(!)
-    for (( i=0; i <= ${#GVM_CANDIDATES}; i++ )); do
+    for (( i=0; i <= ${#GVM_CANDIDATES[*]}; i++ )); do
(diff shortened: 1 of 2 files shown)


danthegoodman:
Sure thing. I can give that a shot. Would you like the test in `src/test/steps/gvm/initialisation_steps.groovy` or a new `src/test/groovy/gvm/specs/InitialisationSpec.groovy`?

marcoVermeulen:
Thanks! It sounds a little more like a Spock spec than a new feature, so `src/test/groovy/gvm/specs/InitialisationSpec.groovy` sounds like the better choice.

Would it be possible to test the current boundary conditions? Maybe have the exact candidate list that you were seeing problems with, and watching it fail first before we adding the fix.

danthegoodman:
@marcoVermeulen I added the test. I wrote it on master, watched it fail, then committed it after my fix and watched it pass.

marcoVermeulen:
Excellent work! Thanks!

PR #364, opened 2015-09-29
Increase curl timeout parameters on sdkman-init.sh

grecovery:
Increase curl *--connect-timeout* and *--max-time* parameters to a bigger value (15/20 seconds instead of 1 second) to restore acceptable startup time of the bash shell.

On my system (ubuntu 14.04.03 x86_64 in Italy) dsn resolution of "api.sdkman.it" took a longer time of the expected behaviour (~8 seconds, view Note 1). The curl utility wait entirely this time despite the --max-time 1 option.

This lead every shell spawned on the system to hangs until curl return. Note that curl fails without actually download anything (the file ~/.sdkman/var/version is not written), so the behaviour is repeated everytime.

* Note 1
```bash
$ time curl "http://api.sdkman.io/app/version" --connect-timeout 1 --max-time 1
curl: (28) Resolving timed out after 1510 milliseconds

real    0m8.076s
user    0m0.004s
sys     0m0.011s
```

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -148,13 +148,13 @@ if [[ ! -f "${SDKMAN_DIR}/var/delay_upgrade" ]]; then
 	touch "${SDKMAN_DIR}/var/delay_upgrade"
 fi
 
-# determine if up to date
+# determine if up to date each 24 hours (60*60*24)
 SDKMAN_VERSION_TOKEN="${SDKMAN_DIR}/var/version"
 if [[ -f "$SDKMAN_VERSION_TOKEN" && -z "$(find "$SDKMAN_VERSION_TOKEN" -mmin +$((60*24)))" ]]; then
     SDKMAN_REMOTE_VERSION=$(cat "$SDKMAN_VERSION_TOKEN")
 
 else
-    SDKMAN_REMOTE_VERSION=$(curl -s "${SDKMAN_SERVICE}/app/version" --connect-timeout 1 --max-time 1)
+    SDKMAN_REMOTE_VERSION=$(curl -s "${SDKMAN_SERVICE}/app/version" --connect-timeout 15 --max-time 15)
     sdkman_force_offline_on_proxy "$SDKMAN_REMOTE_VERSION"
     if [[ -z "$SDKMAN_REMOTE_VERSION" || "$SDKMAN_FORCE_OFFLINE" == 'true' ]]; then
         SDKMAN_REMOTE_VERSION="$SDKMAN_VERSION"
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -148,7 +148,7 @@ if [[ ! -f "${SDKMAN_DIR}/var/delay_upgrade" ]]; then
 	touch "${SDKMAN_DIR}/var/delay_upgrade"
 fi
 
-# determine if up to date each 24 hours (60*60*24)
+# determine if up to date
 SDKMAN_VERSION_TOKEN="${SDKMAN_DIR}/var/version"
 if [[ -f "$SDKMAN_VERSION_TOKEN" && -z "$(find "$SDKMAN_VERSION_TOKEN" -mmin +$((60*24)))" ]]; then
     SDKMAN_REMOTE_VERSION=$(cat "$SDKMAN_VERSION_TOKEN")


GsusRecovery:
An addiction: you can also use **wget** to obtain the desired behaviour (close connection if it lasts longer than 1 second) in this way:
```bash
$ wget -q -T 1 "http://api.sdkman.io/app/version" -O -
```
This fix the delay of shells on startup, but (in case of a slow dsn like mine) it will never successfully download the current sdkman version.

marcoVermeulen:
So it seems to be an issue with curl not honouring it's timeout parameters. How does increasing these timeouts solve this problem? We probably need to look at while curl is not doing what it should instead.

Also, even though wget is a great tool, it's not an option because most operating systems don't come with it pre-installed. We try and make the barrier of entry to sdkman as low as possible.

GsusRecovery:
> So it seems to be an issue with curl not honouring it's timeout parameters. We probably need to look at while curl is not doing what it should instead.

In my system curl is at version 7.35.0. According to this [issue](https://github.com/bagder/curl/issues/383) opened on the github repo of curl, versions after 7.38.0 seems to not suffer anymore from this problem.

I've downloaded the latest version (7.45.0) from github and my test confirm the bugfix:
```bash
# the next command add a 3s delay to all outgoing connections on interface wlan0
$ sudo tc qdisc add dev wlan0 root netem delay 3000ms
$ time ./curl "http://api.sdkman.io/app/version" --connect-timeout 1 --max-time 1
curl: (6) name lookup timed out

real	0m1.030s
user	0m0.014s
sys	0m0.024s
# remove the fake delay
$ sudo tc qdisc del dev wlan0 root netem
```
Note that curl 7.45.0 close the connection after 1 second as expected also with just only one of the two options (--connect-timeout 1 or --max-time 1 equally).

> Also, even though wget is a great tool, it's not an option because most operating systems don't come with it pre-installed. We try and make the barrier of entry to sdkman as low as possible.

We can force curl (every bugged version) to work as expected using timeout:

```bash
$ sudo tc qdisc add dev wlan0 root netem delay 3000ms
$ time timeout -s SIGTERM 1 curl "http://api.sdkman.io/app/version" --connect-timeout 1 --max-time 1 

real	0m1.004s
user	0m0.012s
sys	0m0.006s
$ sudo tc qdisc del dev wlan0 root netem
```
 
> How does increasing these timeouts solve this problem?

We can consider the options we have (current version, relaxed timeout, fixed curl)

For users with a **fast** connection (able to download 'version' file in less than 1 second) it's the same in every case (relaxing timeout does not affect download speed).

If we talk about users with a **slow** connection (downloading 'version' file takes longer than 1 second), the options are these:

* Current version of sdkman-init.sh: everytime a new shell is executed the command line appears only after curl as returned (in my case after ~8 seconds and i use a lot of shells...). curl never actually download 'version' file so the problem persists (really annoying).

* Fixed curl (version >7.38.0 or curl executed via timeout): curl wait 1 second before close the connection (as expected) but the 'version' file is neved downloaded. So the version is never actually checked but every new shell startup is delayed of 1 second for nothing.

* Relaxed timeout (15 seconds or a reasonable longer time): the first time a shell is spawned the startup time is affected by the curl long download operation (the same everytime the 'version' file is older than 24 hours) but all the following shells for the next 24h are not. sdkman version control is actually done.

marcoVermeulen:
The reason we introduced the timeout was so the vast majority of users
wouldn't have to sit waiting for 15 seconds on opening a shell. So simply
increasing the timeout is not an option.

For me it almost feels like we need to make the timeout configurable.
Perhaps something that ignores the timeout settings all together. We can
make this a boolean setting in the config file. Sound like a fair trade?

GsusRecovery:
I understand your point perfectly and the boolean option it's ok for me with an appropriate way to inform the user (during installation?): if all my shells start suddenly to hang how i correlate this with sdkman and the boolean option?

Another possibility may be to automatically evaluate download time of 'version' file during sdkman installation (and update) setting a timeout proportional to that value.

> The reason we introduced the timeout was so the vast majority of users
wouldn't have to sit waiting for 15 seconds on opening a shell.

Consider then to use curl via timeout to actually have a working "timeout" (based on curl version?), with every curl < 7.38.0 the timeout options are worse than no options at all (we have delay and no download).

marcoVermeulen:
The best I can do is add the configuration option and to clearly document
this under the configuration section of the website.

GsusRecovery:
Ok ok :) I hope this thread will be useful as well to others groovy / slow-network users to solve the mistery!

Issue #299, opened 2014-10-19
when running tmux, GVM_INIT prevents setting the PATH again


christoph-frick:
When running tmux, it kills all the `PATH` but keeps the `GVM_INIT`.  i currently workaround this by checking for `TMUX` and reset `GVM_INIT`.  So it would be great, if adding/checking the paths to `PATH` would be done outside the `GVM_INIT` check to ensure a consistant environment.  basically the same as #251

Issue #251, opened 2013-12-07
Problems with screen


eplarson:
I'm using gentoo, gvm 1.3.13, and screen 4.00.03, and it looks like I'm having the same problem as was fixed in issue #155: gvm works fine until I create a screen, then it thinks it's initialized already and doesn't set up the environment.

I was able to work around it with this:

```bash
if [ "$TERM" == "screen" ]; then
        unset GVM_INIT
fi
```

PR #373, opened 2015-10-10
Dynamically reinitialize sdkman

calid:
The current init logic using the SDKMAN_INIT flag is fragile and can break if the PATH is modified/reset, but SDKMAN_INIT is still set to true.  In that case re-sourcing sdkman_init.sh is not sufficient, and it's necessary to also unset SDKMAN_INIT.  I added a test case for this scenario.

This patch changes the init logic to dynamically check if a candidate is on the PATH, instead of relying on storing state in a flag.  If a candidate is not on the PATH it will be prefixed.  This should resolve both #251 and #299.

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -169,23 +169,18 @@ fi
 # Arrays are the only way, but unfortunately zsh arrays are not backward compatible with bash
 # In bash arrays are zero index based, in zsh they are 1 based(!)
 for (( i=0; i <= ${#SDKMAN_CANDIDATES[*]}; i++ )); do
-    # Eliminate empty entries due to incompatibility
-    CANDIDATE_NAME="${SDKMAN_CANDIDATES[${i}]}"
-    CANDIDATE_DIR="${SDKMAN_DIR}/${CANDIDATE_NAME}/current"
-    if [[ -n "${CANDIDATE_NAME}" && -h "${CANDIDATE_DIR}" ]]; then
-        CANDIDATE_HOME_VAR="$(echo ${CANDIDATE_NAME} | tr '[:lower:]' '[:upper:]')_HOME"
-        export $(echo "${CANDIDATE_HOME_VAR}")="$CANDIDATE_DIR"
-
-        # Prefix candidate bin to PATH if it doesn't exist already
-        CANDIDATE_BIN="${CANDIDATE_DIR}/bin"
-        echo "$PATH" | grep -q "${CANDIDATE_BIN}" || PATH="${CANDIDATE_BIN}:${PATH}"
-
-        unset CANDIDATE_HOME_VAR
-    fi
-    unset CANDIDATE_NAME
-    unset CANDIDATE_DIR
+	# Eliminate empty entries due to incompatibility
+	CANDIDATE_NAME="${SDKMAN_CANDIDATES[${i}]}"
+	CANDIDATE_DIR="${SDKMAN_DIR}/${CANDIDATE_NAME}/current"
+	if [[ -n "${CANDIDATE_NAME}" && -h "${CANDIDATE_DIR}" ]]; then
+		CANDIDATE_HOME_VAR="$(echo ${CANDIDATE_NAME} | tr '[:lower:]' '[:upper:]')_HOME"
+		export $(echo "${CANDIDATE_HOME_VAR}")="$CANDIDATE_DIR"
+		echo "$PATH" | grep -q "${CANDIDATE_DIR}" || PATH="${CANDIDATE_DIR}/bin:${PATH}"
+		unset CANDIDATE_HOME_VAR
+	fi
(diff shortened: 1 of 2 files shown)


marcoVermeulen:
Thanks for the PR and for adding a test!

Regarding the content of the PR, I'm all for removing the init flag, in fact I was planning on removing it very soon.

However, would it be possible to explain under what circumstances the path breaks? When will the candidate not be on the path? Do you see it happen often? Hope you don't mind, but I would first like to understand exactly what breaks before merging the PR.

calid:
> However, would it be possible to explain under what circumstances the path breaks? When will the candidate not be on the path?

Any time the path is reinitialized or modified.  Re-sourcing sdkman-init.sh is not sufficient and the SDKMAN_INIT flag needs to be manually unset.

```
calid@xen ~
[✔]▶ which groovy
/home/calid/.sdkman/groovy/current/bin/groovy

calid@xen ~
[✔]▶ tail -3 ~/.bashrc
echo RE-SOURCING SDKMAN
export SDKMAN_DIR="/home/calid/.sdkman"
[[ -s "/home/calid/.sdkman/bin/sdkman-init.sh" ]] && source "/home/calid/.sdkman/bin/sdkman-init.sh"

calid@xen ~
[✔]▶ exec bash --login
RE-SOURCING SDKMAN

calid@xen ~
[✔]▶ which groovy

calid@xen ~
[✘]▶ 

calid@xen ~
[✘]▶ unset SDKMAN_INIT

calid@xen ~
[✔]▶ exec bash --login
RE-SOURCING SDKMAN

calid@xen ~
[✔]▶ which groovy
/home/calid/.sdkman/groovy/current/bin/groovy

calid@xen ~
[✔]▶
```

> Do you see it happen often?

Yes, quite often.  Because I use tmux (which creates a login shell by default) sdkman breaks both as soon as I launch tmux and anytime I create a new tmux window or pane.  This makes sdkman basically unusable for me out of the box.

Both #251 and #299 are also for the same problem with terminal multiplexers.  However as the example above demonstrates, this is not just a terminal multiplexer problem (though it's most annoying there), and manifests anytime the user changes or reinitializes their shell or path.

marcoVermeulen:
Cool, I had a closer look and managed to replicate your issue (thanks again for that test!). On closer inspection, all that is needed is the removal of the check for `SDKMAN_INIT`. Due to changes made recently where we build the path up dynamically each time, the other bit of greping the path and adding if not found is not required.

I've also added a further test that checks we have no _duplicate_ entries. This was the original reason that we introduced the `_INIT` check in the first place.

I'll probably release later in the week. If you can confirm that the problem is fixed, please share your findings here and close the issue.

marcoVermeulen:
Had to rebase the branch so it shows as conflicting, although it has been merged into master.

calid:
Cool, I rebased the branch off of master and re-pushed, so it correctly shows up as merged now (GitHub is so wonderful :) )

calid:
Hmm, so trying to do final a test off a local install, but that doesn't seem to be working for me:

```
$ ./gradlew clean install
Environment is set to: local
:clean
:cleanInstallInit UP-TO-DATE
:prepareScripts
:installInit
:cleanInstallModules UP-TO-DATE
:installModules
:install

BUILD SUCCESSFUL

Total time: 3.245 secs

$ source ~/.sdkman/bin/sdkman-init.sh
bash: /home/dmcali/.sdkman/var/candidates: No such file or directory
find: ‘/home/dmcali/.sdkman/ext’: No such file or directory
touch: cannot touch ‘/home/dmcali/.sdkman/var/delay_upgrade’: No such file or directory
bash: /home/dmcali/.sdkman/var/version: No such file or directory
```

Am I doing it wrong, or is this a bug?

marcoVermeulen:
I think you are installing it with the local environment, so you need to
have a populated db and vertx server running on your machine.

Alternatively you can hook it up to prod by running:
./gradlew -Penv=production install

calid:
Still no luck:
```
$ ./gradlew clean
Environment is set to: local
:clean UP-TO-DATE

BUILD SUCCESSFUL

Total time: 0.851 secs

$ ./gradlew -Penv=production install
Environment is set to: production
:cleanInstallInit UP-TO-DATE
:prepareScripts
:installInit
:cleanInstallModules UP-TO-DATE
:installModules
:install

BUILD SUCCESSFUL

Total time: 1.053 secs
$ source ~/.sdkman/bin/sdkman-init.sh
-bash: /home/calid/.sdkman/var/candidates: No such file or directory
find: `/home/calid/.sdkman/ext': No such file or directory
touch: cannot touch ‘/home/calid/.sdkman/var/delay_upgrade’: No such file or directory
-bash: /home/calid/.sdkman/var/version: No such file or directory
```

calid:
Well regardless, I was able to manually bootstrap things and can confirm the new init logic works

marcoVermeulen:
Perfect, thank you.

PR #382, opened 2015-10-27
Fix "find: xxx: No such file or directory" error when using 'outdated' command

nobeans:
Before, there were empty candidates directories even if it's not installed. Now, there are no directories of candidates which isn't installed. So you should fix it.

```
$ sdk outdated
find: /Users/nobeans/.sdkman/candidates/asciidoctorj: No such file or directory
find: /Users/nobeans/.sdkman/candidates/crash: No such file or directory
find: /Users/nobeans/.sdkman/candidates/griffon: No such file or directory
```


marc0der:
Thanks! I think we might have a similar issue on the current command.

PR #384, opened 2015-11-06
Updates to bash scripts for Windows compatiblity on both Cygwin* and MINGW*

shanman190:
The updates that I have applied relate to how both Cygwin and MINGW both handle symlinks. On a Windows system, symlinks are supported until later versions of Windows (Vista and after), however these symlinks aren't the same as a native Linux symlink in that all of the metadata persists, so instead bash emulators such as Cygwin and MINGW have `ln` create a copy of the folder being targeted instead at the location where the symlink would have been placed. These modifications make the necessary adjustments to allow the deletion and checking for the existance of a folder or a symlink, thus allowing native use of the sdkman library on Windows in either Cygwin and MINGW correctly.

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -153,7 +153,7 @@ for (( i=0; i <= ${#SDKMAN_CANDIDATES[*]}; i++ )); do
 	# Eliminate empty entries due to incompatibility
 	CANDIDATE_NAME="${SDKMAN_CANDIDATES[${i}]}"
 	CANDIDATE_DIR="${SDKMAN_CANDIDATES_DIR}/${CANDIDATE_NAME}/current"
-	if [[ -n "$CANDIDATE_NAME" && -h "$CANDIDATE_DIR" ]]; then
+	if [[ -n "$CANDIDATE_NAME" && ( -h "$CANDIDATE_DIR" || -d "${CANDIDATE_DIR}" ) ]]; then
 		__sdkman_export_candidate_home "$CANDIDATE_NAME" "$CANDIDATE_DIR"
 		__sdkman_prepend_candidate_to_path "$CANDIDATE_DIR"
 	fi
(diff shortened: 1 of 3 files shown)


shanman190:
One thing to note with this implementation is that when utilizing the default version (current), when performing the `sdk list <candidate>` command the current version isn't shown with the `>` indicator. If you perform a `sdk use <candidate>` the indicator is properly shown. This is due to the lack of metadata provided from the symlink.

shanman190:
@bastoker @marc0der I've updated this PR to get the code back up in line with the current master version. From my testing in both Windows (MINGW) and Linux (Bash, Zsh), I have not run into any issues.

mdomenjoud:
:+1: Works perfectly for me on both Windows (MINGW)  and Linux (tested on Bash only)

marc0der:
@shanman190 Thanks for contributing! Unfortunately your branch seems to be breaking the build. In particular the `InstallSpec.groovy` is broken.

If you can adapt your code to make the test pass, I will gladly merge the PR.

shanman190:
@marc0der I've run the tests both on sdkman/sdkman-cli:master and shanman190/sdkman-cli:master and I end up with 3 test failures on both sides, so I'm not 100% for sure if it's just my setup that is causing issues or if there are actually build errors that exist currently on sdkman/sdkman-cli:master.


```
sdkman.cucumber.RunCukeTests > Scenario: A List of Available Candidates can be viewed.Then I see "Candidate List" FAILED

sdkman.cucumber.RunCukeTests > Scenario: A List of Available Candidates can be viewed.classMethod FAILED

sdkman.specs.InstallSpec > should skip installation if SDKMAN has been already installed FAILED
```

marc0der:
I tracked down the failing test to a bad pull-request that got merged earlier. All is working and I've rebased your branch against master and merged it in. I'm also releaseing this to prod as version 3.3.2 so closing this issue.

Thanks for the contribution!

shanman190:
That's awesome to hear! Glad to see this get some Windows love. It's such a useful project when working in bash.

Issue #388, opened 2015-11-16
Provide tests for install script.


felipefzdz:
That script is currently manually tested. This script is core in sdkman so the tests should cover different environments (not only Linux). This issue will cover only some automated testing, I'll leave moving those into Vagrant for future stories.

PR #391, opened 2015-11-23
FEATURE #388 - First spec for install script.

felipefzdz:
Apparently it's not possible to add afterAll hooks into cucumber groovy. That means that as a Wiremock instance is created on startup on 8080 if you want to use Wiremock in non-cucumber tests you have two options: 

1. Creating another Wiremock instance in another port.
2. Reusing previous instance.

First solution is not doable as current setup relies on `prepareTestScripts` gradle task. Install script has the following reference, `SDKMAN_SERVICE="@SDKMAN_SERVICE@"`, that means than even if we set `SDKMAN_SERVICE` to `http://localhost:8081` at `CleanBashEnvBuilder` the value will be overridden by the one set up at `build.gradle`. That's why `CleanBashEnvBuilder` is not even setting up those variables. 

Therefore this PR goes with second proposed option.

### DIFF
--- a/src/test/groovy/sdkman/specs/InstallSpec.groovy
+++ b/src/test/groovy/sdkman/specs/InstallSpec.groovy
@@ -1,22 +1,18 @@
 package sdkman.specs
 
+import java.nio.file.Files
+import java.nio.file.Paths
 import com.github.tomakehurst.wiremock.WireMockServer
 import sdkman.env.BashEnv
 import sdkman.env.CleanBashEnvBuilder
 import sdkman.utils.WireMockServerProvider
 import spock.lang.Specification
 
-import static sdkman.stubs.WebServiceStub.primePlainFileEndpoint
-import static sdkman.stubs.WebServiceStub.primeZipEndpoint
+import static sdkman.stubs.WebServiceStub.primeEndpointWithBinary
 import static sdkman.utils.FilesystemUtils.prepareBaseDir
 
 class InstallSpec extends Specification {
-    final installScriptName = "install.sh"
-    final sdkmanScriptsName = "sdkman-scripts.zip"
     final service = "http://localhost:8080"
-    final testScriptBuildDir = "build/testScripts" as File
-    final distributionsBuildDir = "build/distributions" as File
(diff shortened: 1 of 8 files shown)


src/test/groovy/sdkman/specs/InstallSpec.groovy
```diff
@@ -0,0 +1,58 @@
...
+        when:
+        bash.execute("ls ${sdkmanBaseDir.absolutePath}/.sdkman/bin")
+
+        then:
+        bash.output.contains("sdkman-init.sh")
+    }
+
+    private def primeInstallScriptEndpoint() {
+        def binary = Files.readAllBytes(Paths.get("build/testScripts/install.sh"))
+        primeEndpointWithBinary("/", binary)
```
marc0der: It would probably be better to prime wiremock to with the `primeEndpointWithString()` method, as this needs to return content of type `text/plain`

src/test/groovy/sdkman/stubs/WebServiceStub.groovy
```diff
@@ -6,32 +6,40 @@ class WebServiceStub {
 
     static primeEndpoint(String endpoint, String body) {
         stubFor(get(urlEqualTo(endpoint)).willReturn(
```
marc0der: Perhaps rename this to `primeEndpointWithString`

src/test/groovy/sdkman/stubs/WebServiceStub.groovy
```diff
@@ -6,32 +6,40 @@ class WebServiceStub {
...
+                        .withStatus(200)
+                        .withHeader("Content-Type", "text/plain")
+                        .withBody(body)))
+    }
+
+    static primeEndpointWithBinary(String endpoint, byte[] body) {
+        stubFor(get(urlEqualTo(endpoint)).willReturn(
+                aResponse()
+                        .withStatus(200)
+                        .withHeader("Content-Type", "text/plain")
```
marc0der: Needs to return the content type as `application/octet-stream` if it's binary.

PR #392, opened 2015-11-25
Provide hook to disable GVM alias

flyinprogrammer:
This would be nice so that I can now actually run both sdkman and [gvm](https://github.com/moovweb/gvm) side by side. :smile: 

Awesome job with the total rebrand btw - this is super duper awesome!

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,3 +90,7 @@ To install SDKMAN locally running against your local server, run the following c
 
 	$ ./gradlew install
 	$ source ~/.sdkman/bin/sdkman-init.sh
+
+### Disabling the GVM alias
+
+By default we `alias gvm="sdk"` for backwards compatibility. However, if you'd like to disable this, you can by setting `sdkman_disable_gvm_alias=true` in `${SDKMAN_DIR}/etc/config`
--- a/src/main/bash/install.sh
+++ b/src/main/bash/install.sh
@@ -208,6 +208,7 @@ touch "${sdkman_config_file}"
 echo "sdkman_auto_answer=false" >> "${sdkman_config_file}"
 echo "sdkman_auto_selfupdate=false" >> "${sdkman_config_file}"
 echo "sdkman_insecure_ssl=false" >> "${sdkman_config_file}"
+echo "sdkman_disable_gvm_alias=false" >> "${sdkman_config_file}"
 
 echo "Download script archive..."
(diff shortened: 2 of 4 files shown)


marc0der:
Thank you! Much appreciated!

Just curious: Why you would want to run them side by side? GVM is no longer supported and will be blown away by the first selfupdate.

Also, that alias is there for a purpose: many users feeling nostalgic have asked for the `gvm` command to still work as it did before. It also provides backward compatibility for automated environments like CI.

marc0der:
EDIT: I just saw that you were referring to gvm, the Go Version Manager. In that case we could probably make it configurable instead.

flyinprogrammer:
How about this? 
I force-pushed over the previous commit, now there is an environment variable hook.

marc0der:
Hi Alan, thanks for the great work so far; it's already starting to look a lot better. Currently all configuration happens using config that lives under `~/.sdkman/etc/config`. Please see the [configuration](http://sdkman.io/usage.html#config) section on our site for other such config.

I would expect it to be read from there, with the configuration set to `false` by default. A good name would be something like `sdkman_disable_gvm_alias`.

Hope it's making sense so far!

flyinprogrammer:
Cool stuff - this work? (seems to work locally) I chose to explicitly check for `sdkman_disable_gvm_alias ` to be `true` so that even should `sdkman_disable_gvm_alias` be unset the gvm alias will be used.

I wanted to go write some cucumber tests but when I run `./gradlew test` I get: 

```
1120 tests completed, 284 failed, 693 skipped
:test FAILED
```

Should that many be failing?

marc0der:
Sounds good to me. The cucumber tests are all working, but the build is
simply run with a ./gradlew.

marc0der:
I've gone ahead and implemented this, will roll it out shortly.

flyinprogrammer:
oh ok - i'm sorry i never got you tests - i just got busy with other stuff which is a super lame excuse :-1:

marc0der:
Don't worry about that, it ended up being trivial to change. Thanks for
your contribution anyway, it's much appreciated.

deas:
The alias is set when we set disable to "true"?

Found that slightly unexpected. ;)

```
if [[ "$sdkman_disable_gvm_alias" == "true" ]]; then                                                                                                             
        alias gvm="sdk"                                                                                                                                          
fi
```

marc0der:
Yes, we can fix that ;-)

marc0der:
Fixed, I'll roll it out later today. Serves us right for not starting with a test :-P

PR #416, opened 2016-03-24
Remove quotes to fix a less -R $PAGER

moaxcp:
I ran into an issue on nixos where `$PAGER = less -R`. This break `sdk list` with the error `less -R: command not found`

I wrote a quick test to show this problem on my system.

```
set PAGER="less -R"

function __sdkman_page {
    local PAGER="${PAGER-$(which less)}"

    if [[ -n "$PAGER" ]]; then
        "$@" | $PAGER
    else
        "$@"
    fi
}

__sdkman_page echo "$(curl -s "http://www.google.com/")"
```

For now I can change the PAGER variable locally but I think this should be fixed in general.

Happy Coding!!!

### DIFF
--- a/src/main/bash/sdkman-utils.sh
+++ b/src/main/bash/sdkman-utils.sh
@@ -30,7 +30,7 @@ function __sdkman_page {
     local PAGER="${PAGER-$(which less)}"
 
     if [[ -n "$PAGER" ]]; then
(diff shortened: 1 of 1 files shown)


marc0der:
This has now been addressed by @SamirTalwar in a later PR.

Issue #451, opened 2016-08-08
unsafe bash executions


childnode:
e.g. [sdkman-install.sh](https://github.com/sdkman/sdkman-cli/blob/master/src/main/bash/sdkman-install.sh) uses mkdir that is not a more or less secure bash function but might be overloaded by a simple user function. So, you should update any calls to `mkdir` to s.th. like `/usr/bin/env mkdir` to not execute the user function if it exist but the real mkdir binary.

This expands to many others like: `rm`, etc. too!

marc0der:
Pull requests welcome!

childnode:
#lock /me is working on that

PR #452, opened 2016-08-09
use /usr/bin/env for rm, mv, mkdir, ln

childnode:
fixing https://github.com/sdkman/sdkman-cli/issues/451

> e.g. [sdkman-install.sh](https://github.com/sdkman/sdkman-cli/blob/master/src/main/bash/sdkman-install.sh) uses mkdir that is not a more or less secure bash function but might be overloaded by a simple user function. So, you should update any calls to `mkdir` to s.th. like `/usr/bin/env mkdir` to not execute the user function if it exist but the real mkdir binary.
> 
> This expands to many others like: `rm`, etc. too!


marc0der:
Has this been tested on cygwin with the full suite of tests?

childnode:
> cygwin

pah ;D

no, sorry, I didn't tested this on cygwin (shame on me =) ) but since it's commonly used in scripts, I bet `/usr/bin/env` also works on windows
any doubts?

but perhaps, someone can "quickly" add appveyor.com in addition to travis so we can see this in any build?! :)

grimreaper:
it would be far better to use "command" instead of /usr/bin/env to solve the same problem.

PR #461, opened 2016-08-30
Use https for sdkman setup

Graham42:
Should always use https instead of http

### DIFF
--- a/config.groovy
+++ b/config.groovy
@@ -7,7 +7,7 @@ environments {
 	}
 	production {
 		sdkmanService = 'https://api.sdkman.io'
-		sdkmanBroadcastService = 'http://cast.sdkman.io'
-		sdkmanBrokerService = 'http://broker.sdkman.io'
+		sdkmanBroadcastService = 'https://cast.sdkman.io'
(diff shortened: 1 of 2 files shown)


config.groovy
```diff
@@ -7,7 +7,7 @@ environments {
 	}
 	production {
 		sdkmanService = 'https://api.sdkman.io'
-		sdkmanBroadcastService = 'http://cast.sdkman.io'
-		sdkmanBrokerService = 'http://broker.sdkman.io'
+		sdkmanBroadcastService = 'https://cast.sdkman.io'
```
marc0der: This api does not support https, and will be retired within a week.

config.groovy
```diff
@@ -7,7 +7,7 @@ environments {
...
+		sdkmanBrokerService = 'https://broker.sdkman.io'
```
marc0der: This is never used and will also be retired within a week.

marc0der:
If you could remove the bottom 2 config entries I'll merge your PR.

marc0der:
README fixed.

PR #469, opened 2016-09-09
Remove akka from list of candidates in help

rahulsom:
The help lists akka as one of the candidates, which is weird, because it is not.
This change fixes that by replacing with gradle.

If there is a list of most popular things installed using sdkman, I can replace the list with things from that.


marc0der:
Maybe replace it with sbt instead?

rahulsom:
@marc0der You got it. I've updated the PR with `kotlin`.

marc0der:
Rather meant to replace akka with Kotlin. Gradle is a good one ;-)

rahulsom:
Both are in now.

marc0der:
Done, thank you!

PR #494, opened 2016-12-01
Don't use the output of `which` or `command -v`.

### LINKED ISSUE #493: sdk list command is broken, version 5.1.7+91
Mac OS 10.11.6
Bash

```
$ sdk v
SDKMAN 5.1.7+91

$ sdk list
-bash: less is /usr/bin/less: No such file or directory
-bash: echo: write error: Broken pipe
```

But it worked before this update

I've just checked and found that this issue is related to #411 /CC @SamirTalwar

SamirTalwar:
`which less` has the following output in some versions of Bash:

    less is /usr/bin/less

`command -v less` outputs the path, but unfortunately has non-executable output in the case of aliases:

    $ alias less='my-super-less'
    $ command -v less
    alias less='my-super-less'

In this change, we use the exit code, not the output. (Thanks @halyph.)

Closes #493.

### DIFF
--- a/src/main/bash/sdkman-utils.sh
+++ b/src/main/bash/sdkman-utils.sh
@@ -55,10 +55,10 @@ function __sdkman_secure_curl_with_timeouts {
 }
 
 function __sdkman_page {
-    local PAGER="${PAGER-$(which less)}"
-
     if [[ -n "$PAGER" ]]; then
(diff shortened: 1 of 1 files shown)


marc0der:
Should appear in our beta channel shortly.

PR #497, opened 2016-12-08
Add aria2 for alternate download method.

abcfy2:
`aria2` is a very excellent download tool, than can speed up downloading.
 
Use `aria2` for downloading if exists. `axel` is another download tool, but it cannot handle redirect gracefully.

### DIFF
--- a/src/main/bash/sdkman-utils.sh
+++ b/src/main/bash/sdkman-utils.sh
@@ -105,12 +105,12 @@ function __sdkman_echo_confirm {
 }
 
 function __sdkman_legacy_bash_message {
-    if [[ "$bash_shell" == 'true' && -z "$(bash --version | grep 'GNU bash, version 4..*')" ]]; then
+    if [[ "$bash_shell" == 'true' && "$BASH_VERSINFO" -lt 4 ]]; then
         __sdkman_echo_red "An outdated version of bash was detected on your system!"
         echo ""
         __sdkman_echo_red "We recommend upgrading to bash 4.x, you have:"
         echo ""
-        __sdkman_echo_yellow "  $(bash --version | grep 'GNU bash, version')"
+        __sdkman_echo_yellow "  $(bash --version | head -n1)"
         echo ""
     fi
 }


marc0der:
I really like the idea of using aria2, but have a few remarks about the PR. If you're happy and willing to make the changes I'll merge it.

* currently has no test coverage
* should be driven by a configuration, something like `sdkman_download_tool=curl|aria2`
* would like aria2 to give some progress, not dumping all output to `/dev/null`

abcfy2:
Sorry, I'm not familiar with cucumber, I don't know how to test.

I've already add `sdkman_download_tool` configuration.

marc0der:
Okay, I can help with writing some new tests. However, I can only really merge if all the current tests are passing.

abcfy2:
Test passed. It's my fault that I forgot to set default fallback `sdkman_download_tool`.

src/main/bash/sdkman-utils.sh
```diff
@@ -30,20 +30,42 @@ function __sdkman_secure_curl {
...
 		curl_params="$curl_params --cookie $cookie"
+		aria2_params="$aria2_params --load-cookies=$cookie_file"
 	fi
 
-	curl ${curl_params} "$1"
+	if [[ -z "${sdkman_download_tool}" ]] && which aria2c &>/dev/null; then
+		sdkman_download_tool='aria2'
+		echo sdkman_download_tool=aria2 >> "${SDKMAN_DIR}/etc/config"
+		echo "Find aria2c in PATH. Choose aria2 for downloading automatically."
+		echo "If you want to use curl, set 'sdkman_download_tool=curl' in '${SDKMAN_DIR}/etc/config'"
```
marc0der: Really? Should we be forcing users into using aria2c like this?

I also don't like the verbosity of the output as well as the side-effect of writing config introduced here. This config should be written by the [selfupdate.sh](https://github.com/sdkman/sdkman-candidates-legacy/blob/master/src/main/scripts/selfupdate.sh) script in the legacy candidates service. I also think the default configuration should be `curl`, and we should update the [website's config section](https://github.com/sdkman/sdkman-website/blob/master/site/src/site/pages/usage/config.groovy) to explain the new configuration.

src/main/bash/sdkman-utils.sh
```diff
@@ -30,20 +30,42 @@ function __sdkman_secure_curl {
...
+	elif [[ "${sdkman_download_tool}" == 'aria2' ]] && ! which aria2c &>/dev/null; then
+		sdkman_download_tool='curl'
+		echo "Could not find aria2c in PATH, fallback with curl."
+		echo "To omit this warning, set 'sdkman_download_tool=curl' in '${SDKMAN_DIR}/etc/config'"
```
marc0der: Too verbose, nothing should be printed to stdout here.

src/main/bash/sdkman-utils.sh
```diff
@@ -30,20 +30,42 @@ function __sdkman_secure_curl {
...
+        else
+                sdkman_download_tool='curl'
```
marc0der: Funky spacing going on here...

src/main/bash/sdkman-utils.sh
```diff
@@ -30,20 +30,42 @@ function __sdkman_secure_curl {
...
 		curl_params="$curl_params --insecure"
+		aria2_params="$aria2_params --check-certificate=false"
 	fi
 
 	local cookie_file="${SDKMAN_DIR}/var/cookie"
 
 	if [[ -f "$cookie_file" ]]; then
 	    local cookie=$(cat "$cookie_file")
 		curl_params="$curl_params --cookie $cookie"
+		aria2_params="$aria2_params --load-cookies=$cookie_file"
```
marc0der: Funky spacing here.

abcfy2: I use `vim` to write down my codes. I set using `expandtab` instead of `tab`.

abcfy2:
Typo is easy to fix. You mean that we should consider `curl` as default download tool, without checking `aria2` is existed?

abcfy2:
Hi @marc0der. I've already merged your upstream codes and reformated my codes again. Please checkout. Let me know if you have any questions.

src/main/bash/sdkman-utils.sh
```diff
@@ -30,20 +30,27 @@ function __sdkman_secure_curl {
...
 
 	if [[ -f "$cookie_file" ]]; then
 		local cookie=$(cat "$cookie_file")
 		curl_params="$curl_params --cookie $cookie"
+		aria2_params="$aria2_params --load-cookies=$cookie_file"
 	fi
 
-	curl ${curl_params} "$1"
+	if [[ "${sdkman_download_tool}" == 'aria2' ]]; then
+		aria2c --dir=$(dirname "$2") --out=$(basename "$2") $aria2_params "$1"
```
grimreaper: This might behave poorly if the dirname itself has spaces.

abcfy2: No. `$()` will automatic escape the spaces and other special characters.

marc0der:
No it won't, it merely opens a sub-shell.

src/main/bash/sdkman-utils.sh
grimreaper: Bah, you're right. This is what I get for reviewing code when tired.

grimreaper: Take a look at this:
```
#include <stdio.h>

int main(int argc, char* argv[]) {
	for (int i = 0; i < argc; ++i) {
		printf("%d: %s\n", i, argv[i]);
	}
}
```

```
∴./wordcheck $(echo "a b c")
0: ./wordcheck
1: a
2: b
3: c

∴./wordcheck "$(echo a b c)"
0: ./wordcheck
1: a b c
```
This is slightly difference. With echo it won't matter, but for other commands it might.

abcfy2: Ah... Yes, you are right.

Assign to variable will be OK. But use the variable needs to add `""`. Thanks.

marc0der:
[Add aria2c as download alternative](https://trello.com/c/AjSVbHKv/17-add-aria2c-as-download-alternative)

PR #498, opened 2016-12-08
Fix bug when trying to use curl to get candidate list.

jrpalmer:


marc0der:
Good catch! Thanks for contributing!

PR #502, opened 2016-12-14
Outdated command can update all outdated candidates

pbassiner:
The purpose of this PR is to provide a way to update all the outdated candidates and set them as the default version.

When running `sdk outdated`, with or without a specific candidate, and only if there is any outdated candidate, ask the user for confirmation on updating all of them and setting them as the default, and do so accordingly.

### DIFF
--- a/src/main/bash/sdkman-outdated.sh
+++ b/src/main/bash/sdkman-outdated.sh
@@ -17,7 +17,7 @@
 #
 
 function __sdk_outdated {
-    local all candidates candidate outdated installed_count outdated_count
+    local all candidates candidate outdated installed_count outdated_count outdated_candidates
     if [ -n "$1" ]; then
         all=false
         candidates=$1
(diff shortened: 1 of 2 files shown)


pbassiner:
Sure, it totally makes sense. I've updated the commit and the PR description.

marc0der:
Excellent work! Thanks for your valuable contribution!

marc0der:
@pbassiner I made some minor improvements in what/how is displayed on CLI. Let me know if you agree.

pbassiner:
LGTM. Maybe you could consider adding `auto_answer_outdated` as a config parameter as well and then avoiding the interactive question accordingly.

marc0der:
We already have such a config in the form of `sdkman_auto_answer`. We should probably check for the case that this is present in sdkman-install.sh too. However, our case is slightly different since we only want to fast track the installation when driven from the outdated command.

I'll add a further test for this and include the third case.

PR #504, opened 2016-12-19
Make first installed candidate version automatically the default one

### LINKED ISSUE #442: Make first Version for Candidate default without confirmation
This is similar to #425 but that fix didn't work. This happens in both ZSH and Bash. This is not a new installation of sdkman -- it used to work but I don't know what I changed.

```
bash-3.2$ sdk version
SDKMAN 5.0.0+51
bash-3.2$ sdk use grails 3.0.16

Using grails version 3.0.16 in this shell.
bash-3.2$ which grails
bash-3.2$ grails
bash: grails: command not found
```

Because you asked for this in the other question:

```
bash-3.2$ tree ~/.sdkman/candidates/grails/3.0.16|head -n 20
/Users/mikeh/.sdkman/candidates/grails/3.0.16
├── INSTALL
├── LICENSE
├── bin
│   ├── grails
│   └── grails.bat
├── dist
│   ├── grails-async-3.0.16-javadoc.asc
│   ├── grails-async-3.0.16-javadoc.asc.sha1
│   ├── grails-async-3.0.16-javadoc.jar
│   ├── grails-async-3.0.16-javadoc.jar.sha1
│   ├── grails-async-3.0.16-sources.asc
│   ├── grails-async-3.0.16-sources.asc.sha1
│   ├── grails-async-3.0.16-sources.jar
│   ├── grails-async-3.0.16-sources.jar.sha1
│   ├── grails-async-3.0.16.asc
│   ├── grails-async-3.0.16.asc.sha1
│   ├── grails-async-3.0.16.jar
│   ├── grails-async-3.0.16.jar.sha1
│   ├── grails-bootstrap-3.0.16-javadoc.asc

bash-3.2$ ls -lah ~/.sdkman/candidates/grails/3.0.16/bin
total 40
drwxrwxr-x  4 mikeh  staff   136B Apr 14 13:37 .
drwxr-xr-x  9 mikeh  staff   306B Apr 14 13:38 ..
-rwxrwxr-x  1 mikeh  staff   8.9K Apr 14 13:37 grails
-rw-rw-r--  1 mikeh  staff   6.4K Apr 14 13:37 grails.bat
```

It doesn't seem to matter if I switch versions.

pbassiner:
This fixes #442 
The tests should be reviewed carefully. I had to specify `And the system is bootstrapped` on each test because I needed it to be executed after a `Given the candidate "grails" version "1.3.9" is already installed and default` (checked out the tests of the `default` command for that matter).

### DIFF
--- a/src/test/cucumber/install_candidate.feature
+++ b/src/test/cucumber/install_candidate.feature
@@ -3,62 +3,64 @@ Feature: Install Candidate
   Background:
     Given the internet is reachable
     And an initialised environment
-    And the system is bootstrapped
 
   Scenario: Install a default Candidate
-    Given the candidate "grails" version "2.1.0" is a valid candidate version
+    Given the system is bootstrapped
+    And the candidate "grails" version "2.1.0" is a valid candidate version
     And the default "grails" version is "2.1.0"
-    When I enter "sdk install grails" and answer "Y"
+    When I enter "sdk install grails"
     Then I see "Done installing!"
-    Then the candidate "grails" version "2.1.0" is installed
+    And I do not see "Do you want grails 2.1.0 to be set as default? (Y/n)"
+    And the candidate "grails" version "2.1.0" is installed
 
   Scenario: Install a specific Candidate
+    Given the system is bootstrapped
     And the candidate "grails" version "1.3.9" is available for download
-    When I enter "sdk install grails 1.3.9" and answer "Y"
+    When I enter "sdk install grails 1.3.9"
     Then I see "Done installing!"
-    Then the candidate "grails" version "1.3.9" is installed
+    And I do not see "Do you want grails 2.1.0 to be set as default? (Y/n)"
+    And the candidate "grails" version "1.3.9" is installed
 
   Scenario: Install a Candidate version that does not exist
-    Given the candidate "grails" version "1.4.4" is not available for download
+    Given the system is bootstrapped
+    And the candidate "grails" version "1.4.4" is not available for download
     When I enter "sdk install grails 1.4.4"
     Then I see "Stop! grails 1.4.4 is not available."
 
   Scenario: Install a Candidate version that is already installed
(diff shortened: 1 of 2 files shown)


src/test/cucumber/install_candidate.feature
```diff
@@ -3,62 +3,64 @@ Feature: Install Candidate
...
 
   Scenario: Install a Candidate version that is already installed
-    Given the candidate "grails" version "1.3.9" is available for download
+    Given the system is bootstrapped
+    And the candidate "grails" version "1.3.9" is available for download
     And the candidate "grails" version "1.3.9" is already installed and default
     When I enter "sdk install grails 1.3.9"
     Then I see "Stop! grails 1.3.9 is already installed."
 
-  Scenario: Install a candidate and select to use it
```
marc0der: Don't think this should be deleted, as it test a very valid case.

pbassiner: If adapted to the new feature, I think it would be the same test as `Install a specific Candidate`. That's why I removed it.

pbassiner:
Regarding your first comment, that's what I think this is doing:
* If there's no current version for a candidate, it sets the installed version as default
* Otherwise, it asks for confirmation on setting a new candidate version as default

The last two tests I added (`Install a candidate and select to use it automatically` and `Install a candidate and do not select to use it`) aim to cover both scenarios.

I've already rebased to master and updated the `-n "$CURRENT"` check.

Let me know if I'm missing something regarding the new scenarios and its related tests.

src/test/cucumber/install_candidate.feature
marc0der: It's subtly different because `Install a specific candidate` is the first time a version is installed (so no current default exists). I'm adding the scenario that tests this case too.

marc0der:
I've merged this locally so it doesn't seem to be showing up on github. Thanks for yet another great contribution! Keep them coming!

PR #509, opened 2016-12-27
Add zsh completion

pbassiner:
This PR aims to provide `zsh` completion for the `sdk` command.

Many completions are provided as `zsh` plugins through `oh-my-zsh`, although I find it more convenient to have them in each project. This way, both the project and the completions can evolve accordingly instead of relying on the plugin author to keep the completions up to date.

Other projects such as [hub](https://github.com/github/hub) or [docker-compose](https://github.com/docker/compose) manage `zsh` completions this way.

### DIFF
--- a/etc/README.md
+++ b/etc/README.md
@@ -1,0 +1,23 @@
+# Installation instructions
+
+## zsh
+
+Create a new folder for completions:
+
+```sh
+mkdir -p ~/.zsh/completions
+```
+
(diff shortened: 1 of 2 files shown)


marc0der:
Thanks for yet another great PR! Great work!

pbassiner:
Hi @marc0der,
I've noticed this doesn't make much sense unless we include the `sdk.zsh_completion` file into the sdkman installation files (so it's available under `.sdkman/etc`). WDYT?

PR #523, opened 2017-01-10
Update SDKMAN_VERSION after selfupdate

pbassiner:
After running `selfupdate` the `.sdkman/var/version` file holds the latest version but `SDKMAN_VERSION` environment variable is not updated accordingly. This PR aims to fix that.

I've tried to add a test for it but I've run into a couple of issues:
1. I don't know how `@manual` annotated tests are supposed to be run
2. The kind of test I tried to do I think it's not supported by the current BDD approach:
```
  Scenario: Selfupdate updates version environment variable
    Given an outdated initialised environment
    When I enter "sdk version"
    And I enter "sdk selfupdate"
    And I enter "sdk version"
    Then I see "SDKMAN 4.0.0"
    And I see "Updating SDKMAN..."
    And I see "Successfully upgraded SDKMAN."
    And I see "SDKMAN 5.0.0"
```
Since I'm running more than one command I guess assertions can only access the latest echoed info.

Please let me know if I can add a test somehow.

### DIFF
--- a/src/main/bash/sdkman-selfupdate.sh
+++ b/src/main/bash/sdkman-selfupdate.sh
@@ -30,6 +30,7 @@ function __sdk_selfupdate {
 		export sdkman_debug_mode
 		export sdkman_beta_channel
 		__sdkman_secure_curl "${SDKMAN_LEGACY_API}/selfupdate?beta=${sdkman_beta_channel}" | bash
+		SDKMAN_VERSION=$(cat "$SDKMAN_VERSION_FILE")
 	fi
 	unset SDKMAN_FORCE_SELFUPDATE
 }


marc0der:
I'm not convinced that this is the right thing to do. The reason as follow:

When a selfupdate is performed, the underlying functions potentially change a lot, and are not loaded into the shell automatically. This is why we still report the old version as is, because the loaded program has not changed.

We give the user a clue to simply open a new shell or re-source the `sdkman-init.sh` script, at which point the version will be updated to the latest.

Hope this makes sense!

pbassiner:
I didn't take that into account. I'm closing the PR then. Thanks for the feedback!

marc0der:
Sure, no probs.

pbassiner:
Maybe I'm missing something but even if I open a new shell or re-source `sdkman-init.sh`, I still see the old version. I wonder if this also means that updated functions might not be loaded as well.

marc0der:
This is why we suggest opening a new terminal these days.

pbassiner:
I've tried opening a new terminal as well and it keeps showing the old version.

marc0der:
I think I've seen that behaviour before when using a login shell. Could you try logging out of your session and back in again? Also, what platform are you on?

pbassiner:
I'm on Ubuntu 16.04, using `zsh` along with `terminator`. Logging out and logging back in again shows the proper version.

marc0der:
Okay, so that sounds me a login shell problem, although I'm not sure how this plays with zsh. You could try removing the init script from the .profile and .zprofile files to see if that solves the problem. 

BTW, I'm on Arch linux with zsh and don't have this issue, but have heard other Ubuntu / Mint users speak about having this exact problem.

PR #529, opened 2017-01-24
Fix linking in windows bash environment

xshyamx:
In windows the link command simply creates a copy of the directory and hence the removal of the directory fails without the `-r` flag


src/main/bash/sdkman-path-helpers.sh
```diff
@@ -82,7 +82,7 @@ function __sdkman_link_candidate_version {
 
 	# Change the 'current' symlink for the candidate, hence affecting all shells.
 	if [[ -h "${SDKMAN_CANDIDATES_DIR}/${candidate}/current" || -d "${SDKMAN_CANDIDATES_DIR}/${candidate}/current" ]]; then
-		rm -f "${SDKMAN_CANDIDATES_DIR}/${candidate}/current"
+		rm -fr "${SDKMAN_CANDIDATES_DIR}/${candidate}/current"
```
marc0der: I would consider a recursive forced removal to dangerous.

xshyamx: This is on windows using git-bash

PR #559, opened 2017-04-11
Make upgrade command bash/zsh compatible

pbassiner:
The `upgrade` command doesn't work properly in `zsh` due to the different nature on array assignations.

I couldn't find a more elegant way to do it. This is the only place in the whole codebase where the `SDKMAN_CANDIDATES` array is not directly accessed but assigned to a local variable (in order to reuse the upgrade code for the single candidate vs. all candidates scenarios).

### DIFF
--- a/src/main/bash/sdkman-upgrade.sh
+++ b/src/main/bash/sdkman-upgrade.sh
@@ -23,7 +23,11 @@ function __sdk_upgrade {
 		candidates=$1
 	else
 		all=true
-		candidates=${SDKMAN_CANDIDATES[@]}
+		if [[ "$zsh_shell" == 'true' ]]; then
+			candidates=( ${SDKMAN_CANDIDATES[@]} )
+		else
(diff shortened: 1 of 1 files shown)


marc0der:
Just out of interest, what behaviour are you seeing when using `upgrade` under zsh?

pbassiner:
It's prompting "No candidates are in use" because the for loop `for candidate in ${candidates}; do` is not properly splitting the `candidates` content and `candidate` is assigned all the list (`activator ant ...`) in the first iteration.
This doesn't happen anywhere else because of the zsh-aware code in `sdkman-init.sh` which assigns the candidates to `SDKMAN_CANDIDATES`, and this is the variable used everywhere else.
In `upgrade` we're assigning `SDKMAN_CANDIDATES` to `candidates` with bash-only compliance, and hence the issue in zsh.

marc0der:
Thanks for contributing again!

marc0der:
Released as 5.5.9!

PR #599, opened 2017-08-21
fix for 598

mkrasuski:
for clarity I've tried to make update of cache to be more "transactional" so renamed working SDKMAN_CANDIDATES_CSV to SDKMAN_FRESH_CANDIDATES_CSV during the download and validation

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -98,23 +98,25 @@ fi
 if [[ -z "$sdkman_curl_connect_timeout" ]]; then sdkman_curl_connect_timeout=7; fi
 if [[ -z "$sdkman_curl_max_time" ]]; then sdkman_curl_max_time=10; fi
 
-# fabricate list of candidates
+# maintain candidates cache
 SDKMAN_CANDIDATES_CACHE="${SDKMAN_DIR}/var/candidates"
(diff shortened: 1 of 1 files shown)


marc0der:
Links to #598

marc0der:
Excellent, thanks for looking into that! Would you feel comfortable adding a test for this scenario too? The test you would need to add to is the Spock specification [CandidatesCacheBootstrapSpec](https://github.com/sdkman/sdkman-cli/blob/master/src/test/groovy/sdkman/specs/CandidatesCacheBootstrapSpec.groovy) written in Groovy.

mkrasuski:
I'm not very familiar with this test environment, and all the previous tests were built across candidates cache file. Notice that the problems was located in the logic of using this file i.e. populating bash var SDKMAN_CANDIDATES_CSV with the value of cache file. I don't know how to access this variable from test procedure. Is that possible?

mkrasuski:
In fact all test cases concerned with calculating of the cache are already covered and clue of my change is unconditionally use cache file after its maintenance. Even if we could access SDKMAN_CANDIDATES_CSV from tests unconditional assignment:

SDKMAN_CANDIDATES_CSV="$(cat $SDKMAN_CANDIDATES_CACHE)" 

should be reliable and contents of  $SDKMAN_CANDIDATES_CACHE are tested elsewhere...

marc0der:
As far as I'm concerned, code does not work unless proven by an _explicit_ test. Since this is addressing a bug, a specific test is most certainly required. Merely having it covered by other tests is not sufficient.

If you aren't comfortable writing a test, I can see what I can do over the weekend but can't promise that I'll get around to it.

PR #600, opened 2017-08-25
WIP: proposal for a way to configure dimmer colors

fditolla:
proposal for a way to configure colors that are more readable on light terminals, using a config variable named 'sdkman_colour_style' defaulting to 1 and that when set to 2 uses dimmer colors suitable to white/light colored terminals. (I could not add the variable to the script executed upon installation since is not under configuration here).

### DIFF
--- a/src/main/bash/sdkman-utils.sh
+++ b/src/main/bash/sdkman-utils.sh
@@ -72,7 +72,7 @@ function __sdkman_echo {
 	if [[ "$sdkman_colour_enable" == 'false' ]]; then
 		echo -e "$2"
 	else
-		echo -e "\033[1;$1$2\033[0m"
+		echo -e "\033[${sdkman_colour_style:-1};$1$2\033[0m"
 	fi
 }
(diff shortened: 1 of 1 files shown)


marc0der:
I like the idea, but think it might be more beneficial if the colour be turned off completely. I like the use of the variable, but would probably call it something like sdkman_colour_enable.

fditolla:
sdkman_colour_enable=true is already included in the config file, and it works correctly, this was to change the color to dim them, alternatively you could put a fixed "2" in place of the variable I introduced in the colored echo

marc0der:
Evidently,  it's been a while since I implemented that :-)

Would it be possible to somehow make the config read something like:

    sdkman_colour_style=dark/light

That would be more intuitive for people trying to configure this.

fditolla:
I've committed the change you suggested. 
I've seen the new indication on contributions: apologies for not having posted in gitter before branching and proposing the change.

Issue #624, opened 2018-02-23
sdkman-init.sh optimization and cleanup


jjlin:
We require a conversation to take place in the appropriate Gitter Room before raising a new Github Issue. Please note that **issues will be closed immediately if prior discussion did not take place**. We need to do this to help manage the quality and validity of Issues raised on this project.

Please tick one:
- [x] this is a new Feature Request, a conversation was started in the [Gitter CLI Development Room](https://gitter.im/sdkman/cli-dev).
- [ ] this is a new Bug Report or Issue, a conversation was started in [Gitter User Issues Room](https://gitter.im/sdkman/user-issues).

Please explain the Issue / Feature Request here:

`sdkman-init.sh` has some inefficiencies and unnecessary hacks that should be fairly easy to correct. It's relatively worthwhile to optimize this file because it's run every time a new interactive shell is spawned.

marc0der:
Thanks for the improvements!

PR #625, opened 2018-02-23
sdkman-init.sh: various optimizations and cleanup

jjlin:
In my tests, this reduces execution time by about 50% with `bash`,
and 40% with `zsh`. Most of the speedup is from removing calls to
external commands like `basename` and `cat`.

This addresses Issue #624.

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -17,10 +17,12 @@
 #
 
 SDKMAN_PLATFORM="$(uname)"
-if [[ "$SDKMAN_PLATFORM" == 'Linux' && "$(uname -m)" == 'i686' ]]; then
-    SDKMAN_PLATFORM="${SDKMAN_PLATFORM}32"
-elif [[ "$SDKMAN_PLATFORM" == 'Linux' ]]; then
-    SDKMAN_PLATFORM="${SDKMAN_PLATFORM}64"
+if [[ "$SDKMAN_PLATFORM" == 'Linux' ]]; then
+    if [[ "$(uname -m)" == 'i686' ]]; then
+        SDKMAN_PLATFORM+='32'
+    else
+        SDKMAN_PLATFORM+='64'
+    fi
 fi
 export SDKMAN_PLATFORM
 
@@ -43,11 +45,11 @@ fi
 export SDKMAN_CANDIDATES_DIR="${SDKMAN_DIR}/candidates"
 
 # OS specific support (must be 'true' or 'false').
-cygwin=false;
-darwin=false;
-solaris=false;
-freebsd=false;
-case "$(uname)" in
+cygwin=false
+darwin=false
+solaris=false
+freebsd=false
+case "${SDKMAN_PLATFORM}" in
 	CYGWIN*)
 		cygwin=true
 		;;
@@ -62,8 +64,8 @@ case "$(uname)" in
 esac
 
 # Determine shell
-zsh_shell=false;
-bash_shell=false;
+zsh_shell=false
+bash_shell=false
 
 if [[ -n "$ZSH_VERSION" ]]; then
     zsh_shell=true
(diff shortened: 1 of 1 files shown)


src/main/bash/sdkman-init.sh
```diff
@@ -100,29 +107,22 @@ if [[ -z "$sdkman_curl_max_time" ]]; then sdkman_curl_max_time=10; fi
...
-	unset CANDIDATE_NAME CANDIDATE_DIR
+for candidate_name in "${SDKMAN_CANDIDATES[@]}"; do
+    candidate_dir="${SDKMAN_CANDIDATES_DIR}/${candidate_name}/current"
+    if [[ -h "$candidate_dir" || -d "${candidate_dir}" ]]; then
+        __sdkman_export_candidate_home "$candidate_name" "$candidate_dir"
+        __sdkman_prepend_candidate_to_path "$candidate_dir"
+    fi
 done
-unset i
+unset old_IFS candidate_name candidate_dir
```
jjlin: Renamed these variables to maintain the convention that "local" variables are generally lowercase, and global/exported variables are all uppercase.

src/main/bash/sdkman-init.sh
```diff
@@ -17,10 +17,12 @@
...
-if [[ "$SDKMAN_PLATFORM" == 'Linux' && "$(uname -m)" == 'i686' ]]; then
-    SDKMAN_PLATFORM="${SDKMAN_PLATFORM}32"
-elif [[ "$SDKMAN_PLATFORM" == 'Linux' ]]; then
-    SDKMAN_PLATFORM="${SDKMAN_PLATFORM}64"
+if [[ "$SDKMAN_PLATFORM" == 'Linux' ]]; then
+    if [[ "$(uname -m)" == 'i686' ]]; then
+        SDKMAN_PLATFORM+='32'
+    else
+        SDKMAN_PLATFORM+='64'
+    fi
```
marc0der: Does using nested if statements improve performance, or is this merely a stylistic change?

src/main/bash/sdkman-init.sh
marc0der: I'd prefer leaving the name as is (`OLD_IFS`) since `old_IFS` looks a bit odd. Most examples of this hack on the web have the same naming.

src/main/bash/sdkman-init.sh
```diff
@@ -100,29 +107,22 @@ if [[ -z "$sdkman_curl_max_time" ]]; then sdkman_curl_max_time=10; fi
...
-	CANDIDATE_NAME="${SDKMAN_CANDIDATES[${i}]}"
-	CANDIDATE_DIR="${SDKMAN_CANDIDATES_DIR}/${CANDIDATE_NAME}/current"
-	if [[ -n "$CANDIDATE_NAME" && ( -h "$CANDIDATE_DIR" || -d "${CANDIDATE_DIR}" ) ]]; then
-		__sdkman_export_candidate_home "$CANDIDATE_NAME" "$CANDIDATE_DIR"
-		__sdkman_prepend_candidate_to_path "$CANDIDATE_DIR"
-	fi
-	unset CANDIDATE_NAME CANDIDATE_DIR
+for candidate_name in "${SDKMAN_CANDIDATES[@]}"; do
+    candidate_dir="${SDKMAN_CANDIDATES_DIR}/${candidate_name}/current"
+    if [[ -h "$candidate_dir" || -d "${candidate_dir}" ]]; then
```
marc0der: It's great to see that we can now iterate over an array of strings in ZSH! This wasn't possible a few years back when I wrote the original code. Your approach looks way cleaner.

src/main/bash/sdkman-init.sh
jjlin: It was mainly to avoid repeating the `"$SDKMAN_PLATFORM" == 'Linux'` check.

src/main/bash/sdkman-init.sh
```diff
@@ -72,17 +74,22 @@ else
 fi
 
 # Source sdkman module scripts.
-for f in $(find "${SDKMAN_DIR}/src" -type f -name 'sdkman-*' -exec basename {} \;); do
-	source "${SDKMAN_DIR}/src/${f}"
+old_IFS="$IFS"
```
jjlin: This block and the following can actually be collapsed into one by using a single `find` command, so I'll add that. Even better would've been to just use shell globbing directly, but dealing with the different ways of enabling `nullglob` in bash and zsh is a hassle.

marc0der:
@jjlin Thanks for taking the time and contributing! I'll be releasing your change to the stable environment within a day or two. Thanks again!

jjlin:
Thanks!

PR #639, opened 2018-04-25
shell fix for version sort

rbjorklin:
This should solve the following issues: #256 #73 #104 #268 #323


rbjorklin:
Might also want to add `--version-sort` here: https://github.com/rbjorklin/sdkman-cli/blob/6da40149a133e4d377d39dab45336857d21e9ab2/src/main/bash/sdkman-list.sh#L58

marc0der:
This will unfortunately break all other candidates that have columns of versions. This sorting should be done on the serverside and should be regex based (because all candidates have different version schemes). This will guarantee consistent sorting across all candidates.

I'll be kicking this piece of work off shortly.

Issue #649, opened 2018-05-21
Issue on the bash major version check


grational:
Hi sdkman-cli team, 

I've received a message from the `sdk` tool about my outdated version of bash. It seems strange to me since I'm using a recent version of Lubuntu (16.04). 

It seems a localization problem in the code inside `sdkman-utils.sh` file, line `108`:
```bash
-z "$(bash --version | grep 'GNU bash, version 4..*')"
```
Infact, the actual message given by my `bash --version` is slightly different because of the Italian localization:
```bash
GNU bash, versione 4.3.48(1)-release (x86_64-pc-linux-gnu)
```

jjYBdx4IL:
confirmed, make sure to put  a "export LC_ALL=C" inside a bash script to make it independent of current environment. "export TZ=UTC" is usually a good idea, too.

tmysik:
Same problem here, czech version.

marc0der:
The fix has been merged and is available on the beta channel for testing. If I get the thumbs up I'll release it to stable.

marc0der:
This was released as 5.6.4 and resolves the problem so closing this issue.

PR #650, opened 2018-05-21
Refine bash major version check

grational:
The previous bash version check, `bash --version | grep ... `, may not work when the output is not in English (for example, it didn't work when the output was in Italian). 

For the same reason, I've modified also the line 113, to output the first command output of `bash --version` without checking for the content.

This addresses Issue #649


marc0der:
LGTM, thanks for contributing!

Issue #643, opened 2018-05-03
Uninstall older/unused versions of candidate


Xerus2000:
Please tick one:
- [ ] this is a new Feature Request, a conversation was started in the [Gitter CLI Development Room](https://gitter.im/sdkman/cli-dev).
- [x] this is a new Bug Report or Issue, a conversation was started in [Gitter User Issues Room](https://gitter.im/sdkman/user-issues).

Please explain the Issue / Feature Request here:

When using Gradle, I only ever need the latest version, so i would like to request two related features:

1) Allow me to set a flag for a candidate that tells sdkman that this candidate should always be replaced by the new one instead of keeping the older versions.

2) A command to remove all versions of a cnadidate which are not active.

clehene:
There are two issues. 
- One that, on upgrade old versions are retained without limit
- no reasonable way to remove old versions 

A garbage collection option would be nice. I'd expect it to work just by manually deleting unwanted versions from a directory directly. 
Can someone chime in?

PR #656, opened 2018-06-04
add -r for removing older dir

linychuo:
Hi, I'm very pleasure using this tool, and I'm executing 'sdk upgrade', system prompts the new version of gradle, but when finish installing new version, it can't remove older version, and I found problem about removing, so I add '-r' to resolving this bug. Tks!

### DIFF
--- a/src/main/bash/sdkman-path-helpers.sh
+++ b/src/main/bash/sdkman-path-helpers.sh
@@ -82,7 +82,7 @@ function __sdkman_link_candidate_version {
 
 	# Change the 'current' symlink for the candidate, hence affecting all shells.
 	if [[ -h "${SDKMAN_CANDIDATES_DIR}/${candidate}/current" || -d "${SDKMAN_CANDIDATES_DIR}/${candidate}/current" ]]; then
-		rm -f "${SDKMAN_CANDIDATES_DIR}/${candidate}/current"
+		rm -rf "${SDKMAN_CANDIDATES_DIR}/${candidate}/current"
 	fi
 	ln -s "${SDKMAN_CANDIDATES_DIR}/${candidate}/${version}" "${SDKMAN_CANDIDATES_DIR}/${candidate}/current"
 }


linychuo:
for issue #643

src/main/bash/sdkman-path-helpers.sh
```diff
@@ -82,7 +82,7 @@ function __sdkman_link_candidate_version {
 
 	# Change the 'current' symlink for the candidate, hence affecting all shells.
 	if [[ -h "${SDKMAN_CANDIDATES_DIR}/${candidate}/current" || -d "${SDKMAN_CANDIDATES_DIR}/${candidate}/current" ]]; then
-		rm -f "${SDKMAN_CANDIDATES_DIR}/${candidate}/current"
+		rm -rf "${SDKMAN_CANDIDATES_DIR}/${candidate}/current"
```
antrew: `-r` does not change anything here, because `current` is a symlink.

linychuo: My terminal is MSYS2 on the windows 7, and creating a directory instead of symlink, so '-r' is needed.

marc0der: Sure, but that's a bug very specific to Windows, and forcefully removing a directory that msys has created doesn't solve anything. In fact, it would thus also delete a previously installed candidate! That is an unwanted side effect.

What we really want is msys to handle symlinks in the same way that all other Unix platforms do. Is there perhaps some kind of flag or setting we could use here for cygwin/msys?

linychuo: MSYS2 will creating really directory after executing 'ln -s xx xx', you can see this [issue](https://sourceforge.net/p/msys2/tickets/41/), so as your saying, it might specific deal with thing for MSYS2 on windows

reitzmichnicht:
I would suggest to add a check if current is a symlink or a directory and execute then rm -f or rm -rf

Issue #619, opened 2017-12-31
SDKMAN truncates the version number on list output


McPringle:
We require a conversation to take place in the appropriate Gitter Room before raising a new Github Issue. Please note that **issues will be closed immediately if prior discussion did not take place**. We need to do this to help manage the quality and validity of Issues raised on this project.

Please tick one:
- [ ] this is a new Feature Request, a conversation was started in the [Gitter CLI Development Room](https://gitter.im/sdkman/cli-dev).
- [X] this is a new Bug Report or Issue, a conversation was started in [Gitter User Issues Room](https://gitter.im/sdkman/user-issues).

Please explain the Issue / Feature Request here:

I can reproduce this issue mentioned by @johnnybaloney in the [Gitter User Issues Room](https://gitter.im/sdkman/user-issues) chat with the actual version of SDKMAN. As you can see in the directory listing, there is a version `kop-jdk1.8.0_151`. In the `sdk list java` outout the version number is truncated, the last character is missing: `kop-jdk1.8.0_151` -> `kop-jdk1.8.0_15`

```
$ sdk version
SDKMAN 5.5.15+284

$ ls -l ~/.sdkman/candidates/java/
total 0
drwxr-xr-x  18 mcpringle  staff  576 18 Okt 17:39 8u152-zulu
drwxr-xr-x  14 mcpringle  staff  448  4 Nov 16:12 9.0.1-zulu
lrwxr-xr-x   1 mcpringle  staff   51 12 Dez 18:56 current -> /Users/mcpringle/.sdkman/candidates/java/8u152-zulu
drwxr-xr-x   2 mcpringle  staff   64 31 Dez 11:29 kop-jdk1.8.0_151

$ sdk list java
================================================================================
Available Java Versions
================================================================================
   + kop-jdk1.8.0_15                                                               
   * 9.0.1-zulu                                                                    
     9.0.1-oracle                                                                  
     9.0.0-zulu                                                                    
 > * 8u152-zulu                                                                    
     8u151-oracle                                                                  
     8u144-zulu                                                                    
     8u131-zulu                                                                    
     7u141-zulu                                                                    
     6u65-apple
================================================================================
+ - local version
* - installed
> - currently in use
================================================================================
```

marc0der:
@McPringle @johnnybaloney The reason why you are seeing this long version name truncated is that the table actually holds 4 columns of versions, each column with a max width of 15 chars. An example of this would be the listing for Groovy:
```
================================================================================
[...]

PR #661, opened 2018-07-08
Validate Version on Local Install

beckje01:
For installing local development versions validate the length doesn't exceed 15 char which will end up being truncated in the list view.

This Addresses Issue #619
- [X] a Github Issue was opened for this feature / bug.
- [X] test coverage was added (Cucumber or Spock as appropriate).

### DIFF
--- a/src/main/bash/sdkman-install.sh
+++ b/src/main/bash/sdkman-install.sh
@@ -77,11 +77,21 @@ function __sdkman_install_candidate_version {
 }
 
 function __sdkman_install_local_version {
-	local candidate version folder
+	local candidate version folder version_length version_length_max
+
+	version_length_max=15
 
 	candidate="$1"
 	version="$2"
 	folder="$3"
+
+	#Validate max length of version
+	version_length=${#version}
+
+	if [[ $version_length > $version_length_max ]]; then
+		__sdkman_echo_red "Invalid version! ${version} with length ${version_length} exceeds max of ${version_length_max}!"
+		return 1
+	fi
 
 	mkdir -p "${SDKMAN_CANDIDATES_DIR}/${candidate}"
 
@@ -97,6 +107,7 @@ function __sdkman_install_local_version {
 
 	else
 		__sdkman_echo_red "Invalid path! Refusing to link ${candidate} ${version} to ${folder}."
+		return 1
 	fi
 
 	echo ""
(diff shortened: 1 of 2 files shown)


src/main/bash/sdkman-install.sh
```diff
@@ -77,26 +77,35 @@ function __sdkman_install_candidate_version {
 }
 
 function __sdkman_install_local_version {
-	local candidate version folder
+	local candidate version folder versionLength versionLengthMax
+
+	versionLengthMax=15
```
marc0der: Can you bump this to 21 in order to accommodate the new list view?

src/main/bash/sdkman-install.sh
```diff
@@ -77,26 +77,35 @@ function __sdkman_install_candidate_version {
 }
 
 function __sdkman_install_local_version {
-	local candidate version folder
+	local candidate version folder versionLength versionLengthMax
```
marc0der: It's just a little cosmetic thing but could you make the new variables snake case please?

src/main/bash/sdkman-install.sh
```diff
@@ -77,26 +77,35 @@ function __sdkman_install_candidate_version {
...
+	#Validate max length of version is 15
+	versionLength=${#version}
 
-	# handle relative paths
-	if [[ "$folder" != /* ]]; then
-		folder="$(pwd)/$folder"
-	fi
+	if [[ $versionLength > $versionLengthMax ]]; then
+		__sdkman_echo_red "Invalid version! ${version} with length ${versionLength} exceeds max of ${versionLengthMax}!"
+	else
```
marc0der: It might be worth escaping with a non-zero exit code here rather than relying on an else statement. Here is an example of how we've done it elsewhere: https://github.com/sdkman/sdkman-cli/blob/master/src/main/bash/sdkman-install.sh#L57

src/main/bash/sdkman-install.sh
```diff
@@ -97,6 +107,7 @@ function __sdkman_install_local_version {
 
 	else
 		__sdkman_echo_red "Invalid path! Refusing to link ${candidate} ${version} to ${folder}."
+		return 1
```
beckje01: Added the `return 1` here as well let me know if you want me to follow up with setting up different exit codes depending on the errors.

beckje01:
Let me know if you want this back to 15 missed that on this round of changes but easy to switch

beckje01:
Back to the 15

marc0der:
Thanks for contributing, I'll release it in the same release as the cutover to the new list view.

beckje01:
Thanks for clean codebase

marc0der:
Added version length validation in the vendor-release API too: https://github.com/sdkman/vendor-release/commit/52f5bb80aad9640296fe6f09212bee6f58f32b57

Issue #632, opened 2018-03-22
Add resume support for curl


NightMachinary:
Conversation started at https://github.com/sdkman/sdkman-cli/issues/146#issuecomment-374770482

The current curl implementation is terrible to the point of being unusable on an unstable connection. (Note that it need not be slow, just unstable.)

Using another tool or somehow making the "-c" flag work with curl is much appreciated. Just yesterday I cancelled my Java 9 installation after 6 attempts terminating at around 40-60 percent.

marc0der:
![](https://github.trello.services/images/mini-trello-icon.png) [Add resume support for curl](https://trello.com/c/szRoy2VS/26-add-resume-support-for-curl)

marc0der:
@NightMachinary always feel free to raise a PR for this, we're always happy to receive contributions!

qudongfang:
I am suffering from bad network conditions too.
I'd like to help with this.

elliotweiser:
@qudongfang  There are a few flags that could be helpful, namely `--retry`, `--retry-max-time` and `-C`. The appropriate argument for `-C` would be `-` (should read `-C -`) to indicate that the retry should pick up where it left off. 

After a quick scan, I _think_ this is the only file that would need those options applied:

- src/main/bash/sdkman-utils.sh

Not sure at the moment how to simulate an unreliable network for the sake of testing the change, but should be an interesting endeavor!

marc0der:
We currently already have [several flags](https://sdkman.io/usage#config) to control the behaviour of curl. Nothing is stopping us from adding more flags (like `--retry` or `--retry-max-time` to the list of those already supported. Adding the `-C` flag might be a bit trickier but is still doable.

qudongfang:
Could someone help with closing this issue?

PR #667, opened 2018-08-20
download with retry

qudongfang:
support `curl` options related to retry which might be useful in poor network connection conditions.

This addresses Issue #632


elliotweiser:
@qudongfang Nice!

marc0der:
Thanks for contributing!

PR #668, opened 2018-09-12
Use '>|' to write var/ status files

kwbr:
Setting 'noclobber' in bash or zsh prevents writing to an existing
regular file when using > as redirecting operator. Setting 'noclobber' allows
users of interactive shells to prevent accidential overwriting of
existing files.

bash and zsh allow to use >| as redirection operator so that the
redirection is attempted.

Because the write to the status files is required and not by accident
the >| is used in those cases.

This addresses Issue #669

### DIFF
--- a/src/main/bash/sdkman-availability.sh
+++ b/src/main/bash/sdkman-availability.sh
@@ -85,10 +85,10 @@ function __sdkman_update_broadcast {
 	if [[ "$SDKMAN_AVAILABLE" == "true" && "$broadcast_live_id" != "$broadcast_old_id" && "$COMMAND" != "selfupdate" && "$COMMAND" != "flush" ]]; then
 		mkdir -p "${SDKMAN_DIR}/var"
 
-		echo "$broadcast_live_id" >| "$broadcast_id_file"
+		echo "$broadcast_live_id" | tee "$broadcast_id_file" > /dev/null
 
 		BROADCAST_LIVE_TEXT=$(__sdkman_secure_curl "${SDKMAN_CANDIDATES_API}/broadcast/latest")
-		echo "$BROADCAST_LIVE_TEXT" >| "$broadcast_text_file"
+		echo "$BROADCAST_LIVE_TEXT" | tee "$broadcast_text_file" > /dev/null
 		if [[ "$COMMAND" != "broadcast" ]]; then
 			__sdkman_echo_cyan "$BROADCAST_LIVE_TEXT"
 		fi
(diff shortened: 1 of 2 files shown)


src/main/bash/sdkman-availability.sh
```diff
@@ -85,10 +85,10 @@ function __sdkman_update_broadcast {
 	if [[ "$SDKMAN_AVAILABLE" == "true" && "$broadcast_live_id" != "$broadcast_old_id" && "$COMMAND" != "selfupdate" && "$COMMAND" != "flush" ]]; then
 		mkdir -p "${SDKMAN_DIR}/var"
 
-		echo "$broadcast_live_id" > "$broadcast_id_file"
+		echo "$broadcast_live_id" >| "$broadcast_id_file"
```
marc0der: I do prefer the slightly more verbose syntax using `tee` here. Mind changing that?

src/main/bash/sdkman-availability.sh
```diff
@@ -85,10 +85,10 @@ function __sdkman_update_broadcast {
...
 
 		BROADCAST_LIVE_TEXT=$(__sdkman_secure_curl "${SDKMAN_CANDIDATES_API}/broadcast/latest")
-		echo "$BROADCAST_LIVE_TEXT" > "$broadcast_text_file"
+		echo "$BROADCAST_LIVE_TEXT" >| "$broadcast_text_file"
```
marc0der: As above

src/main/bash/sdkman-cache.sh
```diff
@@ -61,8 +61,8 @@ function ___sdkman_check_version_cache {
 
 		else
 			__sdkman_echo_debug "Overwriting version cache with: $SDKMAN_REMOTE_VERSION"
-			echo "${SDKMAN_REMOTE_VERSION}" > "$version_file"
+			echo "${SDKMAN_REMOTE_VERSION}" >| "$version_file"
```
marc0der: And here.

marc0der:
Of course, the bash scripts used in sdkman aren't all in the bash client. Some of it is streamed from the serverside by the [sdkman-hooks](https://github.com/sdkman/sdkman-hooks/blob/master/app/views/selfupdate.scala.txt) service. We would probably need to change this too so that it is consistent throughout.

src/main/bash/sdkman-availability.sh
kwbr: Yes, I will do

marc0der:
Thanks for contributing!

rgoldberg:
I know that this issue PR is old and already merged, but I was just wondering: why did you switch writing the files from using `>|` to using `tee`?

`>|` worked on zsh 5.8, zsh 5.7.1, bash 5.0.16(1)-release & bash 3.2.57(1)-release on macOS.

Issue #673, opened 2018-10-16
It should be possible to use system artifact instead of sdkman's


cardil:
Please tick one:
- [ ] this is a new Feature Request, a conversation was started in the [Gitter CLI Development Room](https://gitter.im/sdkman/cli-dev).
- [x] this is a new Bug Report or Issue, a conversation was started in [Gitter User Issues Room](https://gitter.im/sdkman/user-issues).

Please explain the Issue / Feature Request here:

It's currently impossible, after using sdkman, to restore Java to use default installed in operation system. It should be possible to do that, even temporarly.

It's done this way by multiple tool version managers like:

 * RVM (https://rvm.io/rvm/basics)
 * NVM (https://github.com/creationix/nvm#system-version-of-node)
 * pyenv (https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-global)

I think sdkman could simply remove current binding for a tool with special command, removing `current` symlink for example:

`sdk use java system`

or

`sdk unuse java`

marc0der:
I've got to say that I _really_ like the idea of introducing the new `unuse` command here. I'd expect it to work as an opposing force to the `use` command ie. _it only affect the current shell_.

In addition I would introduce another new command as counterpart to the `default` command, and call that `system`. ie. _affects all subsequent shells that are opened_.

So, this is what you would have:

    $ sdk system java     # use system java in all shells
    $ sdk default java    # use sdkman java in all shells

    $ sdk unuse java      # use the system java only in this current shell
    $ sdk use java        # use sdkman java in the current shell

Does this sound like an acceptable approach to everyone?

cardil:
I must admit I rather like the other approach:

```
$ sdk use java 8-zulu   # use sdkman java 8 in shell
$ sdk use java system # use os java in shell

$ sdk default java 11-zulu # use sdkman java 8 in all future shells
$ sdk default java system # use system java in all future shells
```

I think I just more used to that approach in other tools like rvm or pyenv. It's also more in line with other commands of sdkman. I think it will be easier to remember for users.

But adding two new commands is just fine as well.

It's basically the choosing between:

* adding 2 new commands: `unused` and `system`
* adding virtual version `system`

vmj:
+0 for virtual version.

marc0der:
@vmj did you mean +1? 😀

vmj:
[...]

PR #684, opened 2019-01-21
Update help script with local installation path argument.

hickst:
Updated help script with 4th `install` argument; an optional path to a local installation (for adding `local` versions). This resulted from a conversation held in issue #673.


marc0der:
I've merged and will make the small change when I have some time. Thanks for contributing!

PR #718, opened 2019-11-06
Reset SDKMAN_DIR if it has wrong owner

reitzig:
When `su`-ing to another user, environment variables carry over. 
As a result, `sdk` will point to the installation of the original user, 
not the one running the current shell.

Thus: Reset `$SDKMAN_DIR` if the directory it points at
belongs to a user other than the current one.

(Sorry, I didn't have the time to fulfill the prerequisites. I empathize, but that doesn't change my time prioritization.)

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -25,7 +25,7 @@ if [ -z "$SDKMAN_CANDIDATES_API" ]; then
 	export SDKMAN_CANDIDATES_API="@SDKMAN_CANDIDATES_API@"
 fi
 
-if [ -z "$SDKMAN_DIR" ]; then
+if [ -z "$SDKMAN_DIR" ] || [ $(ls -ld "${SDKMAN_DIR}" | awk '{ print $3 }') != $(whoami) ]; then
 	export SDKMAN_DIR="$HOME/.sdkman"
 fi
 


src/main/bash/sdkman-init.sh
```diff
@@ -25,7 +25,7 @@ if [ -z "$SDKMAN_CANDIDATES_API" ]; then
 	export SDKMAN_CANDIDATES_API="@SDKMAN_CANDIDATES_API@"
 fi
 
-if [ -z "$SDKMAN_DIR" ]; then
+if [ -z "$SDKMAN_DIR" ] || [ $(ls -ld "${SDKMAN_DIR}" | awk '{ print $3 }') != $(whoami) ]; then
```
marc0der: Although I do see the point, the implementation does feel a bit hacky to me.
Can we assign this expression to a clearly named variable? Can we unset said variable at the end of the script? Can we collapse the two conditional blocks into one?

marc0der: Ah, and of course, please refrain from using `awk` as this forces a hard dependency that will often not be met on very basic systems.

reitzig: > Ah, and of course, please refrain from using `awk` as this forces a hard dependency that will often not be met on very basic systems.

Which systems are that? The smallest I know comes with (an) awk:

```
$ docker run --rm alpine awk -version
BusyBox v1.30.1 (2019-06-12 17:51:55 UTC) multi-call binary.
```

Of course, BusyBox binaries are quite often not compatible with GNU ones... :/

That said, which dependencies _are_ you comfortable with? For instance, I [used to use](https://github.com/reitzig/sdkman-for-fish/commit/12f7d3de776402217e452900126145db564087a2#diff-37c089de242f7af37405295c31930375L80) the innocuous `stat` but it turns out that macOS and Ubuntu come with incompatible versions (FreeBSD-ish vs GNU-ish). `ls` + `awk` was the best I could come up with -- I'm open for suggestions!

marc0der: The `cut` command is preferable as it is packaged as part of gnu coreutils. You could try something like this:

```
ls -ld "${SDKMAN_DIR}" | cut -d" " -f3
```

reitzig: > Although I do see the point, the implementation does feel a bit hacky to me.

Do you have a proposal? I felt it's a pretty straight-forward implementation for "do we point at this user's installation?". Of course, it will fail with system installations, but IIRC that's not a use-case you support?

> Can we assign this expression to a clearly named variable? Can we unset said variable at the end of the script? 

Sure!

> Can we collapse the two conditional blocks into one?

What do you mean? `[ -z ... ] || [ ... ]` into `[ ... ]`? No, I don't think so; `[ 1 -eq 1 || 1 -eq 2 ]` doesn't parse. `[[ 1 -eq 1 || 1 -eq 2 ]]` would work, but you use single brackets everywhere else?

marc0der: Actually, yes I do think so. We use double square brackets throughout the sdkman codebase when dealing with complex expressions.

reitzig: > The `cut` command is preferable as it is packaged as part of gnu coreutils. You could try something like this:
> 
> ```
> ls -ld "${SDKMAN_DIR}" | cut -d" " -f3
> ```

This _does_ fail on the current Alpine (BusyBox v1.30.1). :/

Then, maybe `stat` wrapped away in a function and distinguish flavors there? _That's_ in `coreutils`, after all.

reitzig:
It occurs to me that `SDKMAN_VERSION` is similarly affected. Is there a reason against not setting it to `$(cat ${SDKMAN_DIR}/var/version)`?

smac89:
To run a command as another user, don't use `su`, use `sudo` instead. By default, sudo will clear the current user's environment variables before starting the command.

```
sudo --user=<user> <cmd...>
```

PR #719, opened 2019-11-06
Update SDKMAN_VERSION after selfupdate

reitzig:
_Before:_ `sdk version && sdk selfupdate && sdk version` shows the same version twice despite a newer one having been installed in between.

_After:_ Second call prints the new version.


marc0der:
This is not the desired behaviour, the reason being that your shell still has the old version sourced. Until you source the `sdkman-init.sh` explicitly or open a new terminal you are still running the old version.

Please stick to the [contributor guidelines](https://github.com/sdkman/sdkman-cli/blob/master/CONTRIBUTING.md) as stipulated prior to opening new issues and/or PRs.

Issue #683, opened 2019-01-10
Per project sdk configuration


ctranxuan:
Please tick one:
- [X] this is a new Feature Request, a conversation was started in the [Gitter CLI Development Room](https://gitter.im/sdkman/cli-dev).
- [ ] this is a new Bug Report or Issue, a conversation was started in [Gitter User Issues Room](https://gitter.im/sdkman/user-issues).

Please explain the Issue / Feature Request here:

Assume you have an environment managed with sdkman such as Java 11 by default, Maven 3.6.0 by default, etc.
But you have a project that requires Java 8 and Maven 3.5.3. It would be convenient if every-time you enter to the directory of the project, all your environment is setup with Java 8 and Maven 3.5.3 without having to settle it manually through multiple `sdk` command lines.

jenv.io has such nice feature (https://github.com/linux-china/jenv#jenvrc-support): it's addressed through a `jenvrc` file that contains the version of libraries we wish to customize. Then, you call a command `jenv init` that settles the right versions.

**Proposed feature:** being able to declare a `sdkmanrc` into a directory that would contain tools versions and a command `sdk init` that would set for the tools the versions declared in the file `sdkmanrc`

**Example:**
By default, we have java 11 and maven 3.6.0.
In the directory `my-project`, we have a `sdkmanrc` files that contains:

```
java=1.8.0_112
maven=3.5.3
```

Then doing:

```
cd my-project
sdk init
```

would setup the environment to java 1.8.0_112 and maven 3.5.3.

marc0der:
![](https://github.trello.services/images/mini-trello-icon.png) [Per project sdk configuration](https://trello.com/c/J3TlESvY/60-per-project-sdk-configuration)

marc0der:
I would be keen on taking this a step further. How about we introduce a new top level `env` command. When you do an `sdk env init`, it will _create_ an `.sdkman` file in the current directory with all the current versions set in that shell.

Say you opened a shell and then used `sdk use` to set a different version of java, this new version of java (along with any other candidates and versions installed with sdkman already) will be written in this file. To recall the versions, a simple `sdk env` (or `sdk env load`) would `use` all of those in the current shell.

A proposal was also made by @EugenMayer that we add support for language specific descriptors like `.java-version` or `.gradle-version` etc. This should be simple to do, although the preferred method would still be to use an `.sdkman` file instead.

lalloni:
[...]

PR #730, opened 2020-03-08
Add sdk env command to support per-project configuration

helpermethod:

This addresses Issue #683

### DIFF
--- a/src/main/bash/sdkman-env.sh
+++ b/src/main/bash/sdkman-env.sh
@@ -29,7 +29,7 @@ function __sdk_env {
 
   while IFS= read -r line || [[ -n $line ]]; do
     if [[ ! $line =~ ^([[:lower:]]+)=(.+)$ ]]; then
-      __sdkman_echo_red "${sdkmanrc}:${line_number}: Invalid candidate entry! Expected 'candidate=version' but found '$line'"
+      __sdkman_echo_red "${sdkmanrc}:${line_number}: Invalid candidate entry format! Expected 'candidate=version' but found '$line'"
 
       return 1
     fi
--- a/src/main/bash/sdkman-env.sh
+++ b/src/main/bash/sdkman-env.sh
@@ -25,12 +25,7 @@ function __sdk_env {
 
   local candidate version
 
-  while IFS='=' read -r candidate version; do
+  while IFS='=' read -r candidate version || [[ -n $candidate ]]; do
     __sdk_use "$candidate" "$version"
   done < .sdkmanrc
-
-  # if the last line is not terminated by a newline character, `candidate` and
-  # `version` will be set but `read` will return a non-zero status code,
-  # resulting in the loop body not being executed
-  [[ -n $candidate ]] && __sdk_use "$candidate" "$version"
 }
--- a/src/main/bash/sdkman-env.sh
+++ b/src/main/bash/sdkman-env.sh
@@ -34,10 +34,7 @@ function __sdk_env {
       return 1
     fi
 
-    local candidate=${BASH_REMATCH[1]}
-    local version=${BASH_REMATCH[2]}
-
-    __sdk_use "$candidate" "$version"
+    __sdk_use "${BASH_REMATCH[1]}" "${BASH_REMATCH[2]}"
 
     ((line_number++))
   done < "$sdkmanrc"
--- a/src/main/bash/sdkman-env.sh
+++ b/src/main/bash/sdkman-env.sh
@@ -34,7 +34,7 @@ function __sdk_env {
       return 1
     fi
 
-    __sdk_use "${line%*=}" "${line#=*}"
+    __sdk_use "${line%=*}" "${line#*=}"
 
     ((line_number++))
   done < "$sdkmanrc"
--- a/src/main/bash/sdkman-env.sh
+++ b/src/main/bash/sdkman-env.sh
@@ -28,13 +28,13 @@ function __sdk_env {
   local line_number=0
 
   while IFS= read -r line || [[ -n $line ]]; do
-    if [[ ! $line =~ ^([[:lower:]]+)=(.+)$ ]]; then
+    if [[ ! $line =~ ^[[:lower:]]+=.+$ ]]; then
       __sdkman_echo_red "${sdkmanrc}:${line_number}: Invalid candidate format! Expected 'candidate=version' but found '$line'"
 
       return 1
     fi
 
-    __sdk_use "${BASH_REMATCH[1]}" "${BASH_REMATCH[2]}"
+    __sdk_use "${line%*=}" "${line#=*}"
 
     ((line_number++))
   done < "$sdkmanrc"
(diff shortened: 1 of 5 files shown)


src/test/cucumber/per_project_configuration.feature
```diff
@@ -0,0 +1,23 @@
+Feature: Per-project configuration
+
+  Background:
+    Given the internet is reachable
+    And an initialised environment
+
+  Scenario: Env without providing an .sdkmanrc file
```
marc0der: _(the one where)_ The env command is issued without an sdkman project configuration present

I like framing these scenario headlines in the way that Friends episodes are titled :smile:

src/test/cucumber/per_project_configuration.feature
```diff
@@ -0,0 +1,23 @@
...
+    Given the system is bootstrapped
+    When I enter "sdk env"
+    Then I see ".sdkmanrc not found."
```
marc0der: This leads me to think about how we can generate one of these descriptors in the future. It would be nice to eventually do `sdk env generate` or something like that to create a basic descriptor. For now this is perfect though.

src/main/bash/sdkman-env.sh
```diff
@@ -0,0 +1,36 @@
...
+    __sdkman_echo_red '.sdkmanrc not found.'
+
+    return 1
+  fi
+
+  local candidate version
+
+  while IFS='=' read -r candidate version; do
+    __sdk_use "$candidate" "$version"
+  done < .sdkmanrc
```
marc0der: Can we extract `.sdkmanrc` to a constant? It referenced twice so far.

src/main/bash/sdkman-env.sh
```diff
@@ -0,0 +1,36 @@
...
+
+  # if the last line is not terminated by a newline character, `candidate` and
+  # `version` will be set but `read` will return a non-zero status code,
+  # resulting in the loop body not being executed
```
marc0der: Lets try finding a cleaner solution for this if we can. And maybe some tests to cover this would also be good. Perhaps if we take responsible for generating this file with an `sdk env` command extension we wouldn't need to be so defensive.

For now I would suggest failing hard with a proper error message rather than trying to brute-force the last line. wdyt?

src/test/groovy/sdkman/support/UnixUtils.groovy
```diff
@@ -8,7 +8,7 @@ class UnixUtils {
     static asSdkmanPlatform(platform) {
         def result
         switch (platform) {
-            case "Mac OSX":
+            case "Mac OS X":
```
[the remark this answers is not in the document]
helpermethod: Yes, sorry, that was pure laziness. I need I should have done that in a separate commit.

src/main/bash/sdkman-env.sh
helpermethod: I'm more of a rule of three kind a guy but I guess it doesn't hurt.

src/main/bash/sdkman-env.sh
helpermethod: The cleaner solution would be to not use Bash, but that is out of the question. 
Other languages, like AWK, Groovy, you name it, handle this case just fine.

The idea is taken from http://mywiki.wooledge.org/BashFAQ/001#CA-47120ef7c8534fbb8d04287560707aee3db999b5_1.

Alternatively we could rewrite the expression to

```sh
while IFS='=' read -r candidate version || [[ -n $candidate ]]; do
  __sdk_use "$candidate" "$version"
done < .sdkmanrc
```
Same logic, same flaws, though 😄.

I think the check/hack here is better than throwing an error because people will write/modify the file by hand, even when providing a way to do this with `sdkman`. This prevents the script from breaking in that case.

Also, it seems to be impossible to find out which error led read to fail.

helpermethod: Also I was thinking, if we should not also

* strip leading / trailing whitespace
* allow comments

Too much?

helpermethod: In any case we should write tests for each case.

src/test/cucumber/per_project_configuration.feature
helpermethod: Maybe something like `sdk env add groovy 2.5.9`?

src/main/bash/sdkman-env.sh
helpermethod: Should we still handle the valid entries?

src/main/bash/sdkman-env.sh
marc0der: Me three :smile:, but in this case the string literal could be error prone.

src/main/bash/sdkman-env.sh
marc0der: I would refuse to process the file until corrected. wdyt?

    No newline at the end of the file. The .sdkmanrc will be ignored until corrected.

Or something like this.

helpermethod: I am a little torn on this. Most tools handle that case (missing newline) just fine (even older ones like AWK).

My proposed fix (even if quite ugly) would work in those cases and people would be happy. Otherwise you'd may see questions regarding the missing newline pop up on StackOverflow or GitHub (StackOverflow is actually full of questions regarding similar warnings / errors).

helpermethod: See https://mywiki.wooledge.org/BashFAQ/001#My_text_files_are_broken.21__They_lack_their_final_newlines.21

which descrives the fix.

helpermethod:
Also https://github.com/hashicorp/hcl2/issues/18 why it may make sense to handle that case transparently.

src/main/bash/sdkman-env.sh
marc0der: There do seem to be other ways of getting around the problem. See here:
https://stackoverflow.com/questions/1521462/looping-through-the-content-of-a-file-in-bash

In particular, this approach looks promising:

```
cat peptides.txt | while read line || [[ -n $line ]];
do
   # do something with $line here
done
```

helpermethod: Changed it to 

```sh
  while IFS='=' read -r candidate version || [[ -n "$candidate" ]]; do
    __sdk_use "$candidate" "$version"
  done < "$sdkmanrc"
```

which is basically the solution you proposed, excluding the `cat .sdkmanrc` (I use a redirection instead as it performs better than a `cat` followed by a `|`).

marc0der:
@helpermethod I'm hesitant to merge this based on the comment by @djalan. Have you been able to replicate the issue on ZSH?

helpermethod:
@marcoder Where can I find the comment?

helpermethod:
Hi @djalan,

I've replaced the `BASH_REMATCH`with an equivalent parameter expansion. This should also work in ZSH.

hermitmaster:
It would be awesome to see this functionality merged it. It would spare me from having to install jenv.

marc0der:
Thanks for the awesome work @helpermethod. It is now available on the beta channel and I'll do a release later this week.

hermitmaster:
@marc0der Any word on the release?

marc0der:
@hermitmaster we need to complete this feature and add site docs before releasing. In the meanwhile you can get all the goodness in the beta channel where it is deployed.

Issue #722, opened 2019-12-05
Provide a way to get the home directory of an installed sdk candidate version


bes:
Please tick one:
- [x] this is a new Feature Request, a conversation was started in the [Slack CLI Development Channel](https://sdkman.slack.com/app_redirect?channel=cli-development).
- [ ] this is a new Bug Report or Issue, a conversation was started in [Slack User Issues Channel](https://sdkman.slack.com/app_redirect?channel=user-issues).

Please explain the Issue / Feature Request here:

Expose a new command `sdk home <candidate> <version>` to be able to get e.g. a `JAVA_HOME` path to be able to execute one-off command line commands. 

Example

```
# Need Java 11 to build something
sdk use java 11.0.5.hs-adpt
# But there is this one tool that requires Java 8,
# and now you can run it like this without switching sdk
JAVA_HOME=`sdk home java 8.0.232.hs-adpt` my_java_8_command
```

Thank you for considering my feature request.

marc0der:
I'm not opposed to adding this command as I can see that it could be useful.

I do have one silly question though: why not run this other tool in another terminal that `use`s Java 8?

bes:
Good question. My use case for this is I have an alias for a java command that I would like to run interactively from time to time. Maybe it can be solved with a script and a subshell just as well.

cwensel:
fwiw, it's quite common in the Java ecosystem for wrapper shell scripts to rely on a $JAVA_HOME env var, for good or bad (Gradle, Hadoop, Jetty, etc).

The Java SDK analog to this is `/usr/libexec/java_home` which simplifies swapping JDK versions during testing and development. But it only works with Oracle supplied SDKs from what I can tell. 

Leveraging sdkman to be the source of truth of installed versions and paths would be very handy.

rgoldberg:
Should this issue be closed, since a PR was merged for it?

PR #732, opened 2020-03-09
Add a command to print the home path of an installed SDK

bes:

This addresses Issue #722


src/test/cucumber/home.feature
```diff
@@ -0,0 +1,38 @@
+Feature: Print home path
+
+  Background:
+    Given the internet is reachable
+    And an initialised environment without debug prints
```
marc0der: Why without debug prints?

src/main/bash/sdkman-main.sh
```diff
@@ -44,8 +44,21 @@ function sdk {
...
+		ho)
+			COMMAND="home";;
 	esac
 
+	#
+	# Print home directory - special case
+	# Printing the home directory does not require reaching out to the internet.
+	# Printing the home directory also has the hard requirment of not printing
+	# anything else unless it is an actual error
+	#
```
marc0der: Let's drop comments from the source and keep it inside the feature file.

src/main/bash/sdkman-main.sh
```diff
@@ -44,8 +44,21 @@ function sdk {
 			COMMAND="default";;
 		b)
 			COMMAND="broadcast";;
+		ho)
```
marc0der: I'd make this `h` and drop the `help` mnemonic.

src/main/bash/sdkman-help.sh
```diff
@@ -27,6 +27,7 @@ function __sdk_help {
 	__sdkman_echo_no_colour "       list      or ls   [candidate]"
 	__sdkman_echo_no_colour "       use       or u    <candidate> <version>"
 	__sdkman_echo_no_colour "       default   or d    <candidate> [version]"
+	__sdkman_echo_no_colour "       home      or ho   <candidate> <version>"
```
marc0der: Same as before, make this `h` and drop `help` mnemonic.

src/test/cucumber/home.feature
bes: Because then I can just do .trim() on the output and compare it to the path I expect. Otherwise I would need to sanitize the output by first removing all the debug prints that occur. My thought was that this was a more robust way of proving that the output I expect is correct. If another error print is added for some reason (not debug) then the test will catch that right away.

src/main/bash/sdkman-main.sh
bes: done

src/main/bash/sdkman-main.sh
bes: done

src/main/bash/sdkman-help.sh
bes: done

marc0der:
It looks like we have a failing test. Once you get that green I'm happy to merge this.

bbzg:
It's fixed now. The reason I didn't catch this was because I ran the `docker run --rm -it sdkman-cli/gradle test` command without reflecting on the fact that the local source was not mounted and the image had not been rebuilt recently.

I'm not submitting a PR but I think the README file could use a small tweak, namely adding `docker run --rm -it -v $PWD:/usr/src/app sdkman-cli/gradle test` as a better alternative than `docker run --rm -it -v $PWD:/usr/src/app -v $HOME/.gradle:/root/.gradle sdkman-cli/gradle test`. Why? Because something "bricked" my sdkman dev env, I had to completely wipe everything and kill all the gradle daemons to get it working again. I guess because I Ctrl+C'd at exactly the wrong moment at some point, but don't quote me on that... Anyway on my machine the time difference is negligble so I prefer than one since it is very clean and does not mess with my local machine except for mounting the sources.

bes:
It's fixed now. The reason I didn't catch this was because I ran the `docker run --rm -it sdkman-cli/gradle test` command without reflecting on the fact that the local source was not mounted and the image had not been rebuilt recently.

I'm not submitting a PR but I think the README file could use a small tweak, namely adding `docker run --rm -it -v $PWD:/usr/src/app sdkman-cli/gradle test` as a better alternative than `docker run --rm -it -v $PWD:/usr/src/app -v $HOME/.gradle:/root/.gradle sdkman-cli/gradle test`. Why? Because something "bricked" my sdkman dev env, I had to completely wipe everything and kill all the gradle daemons to get it working again. I guess because I Ctrl+C'd at exactly the wrong moment at some point, but don't quote me on that... Anyway on my machine the time difference is negligble so I prefer than one since it is very clean and does not mess with my local machine except for mounting the sources.

marc0der:
Thanks for the excellent work! I'm going to play with it on the beta env over the weekend, then will release and announce it shortly after.

PR #734, opened 2020-03-27
Combined PR

rgoldberg:
This PR combines multiple bug fixes, improvements, cleanups, etc.

I mentioned it on the Slack channel.

I haven't directly discussed all changes in this PR; it will be easier for you to see them by looking at the various commits.

The main changes were:

1. fixed multiple bugs that made tests fail on macOS
2. upgraded Gradle to 6.3, upgraded many dependencies, used Gradle Kotlin DSL instead of Groovy DSL
3. added API functions & validation functions with standardized interfaces
4. improved encapsulation, with command functions validating their arguments themselves instead of `_sdk` validating them
5. improved (standardized, colorized, and output in more places) usage instructions & error messages
6. fixed bug where candidate cache won't be updated if the same non-zero amount of candidates were added and removed
7. fixed an incorrect echo in `sdk install`
8. fixed bug where unassigned `fresh_candidates` was used
9. `__sdkman_link_candidate_version` now uses `rm -rf` instead of `rm -f` on `current` symlink to directory / directory to handle cases such as git bash on Windows without symlinks enabled copying a directory instead of creating a symlink when `ln` runs
10. simplified various code, e.g.:
   - use builtins instead of external commands
   - use `grep -i` instead of using `tr` to convert `grep` input case
   - use already calculated values instead of re-calculating them
   - use code that works in both bash & zsh instead of using a switch for different code for each shell
   - removed unnecessary code
11. use return codes instead of grepping stdout to return info from functions
12. future-proofed code, e.g.:
   - pass all arguments from `_sdk` to commands, instead of just 4, in case commands support more than 4 arguments in future
   - detect http failures using status codes instead of grepping for `html` (which now allows valid responses to contain `html`)
13. redid mutually-exclusive boolean variables as single string variables, e.g.:
   - `shell_name` replaced both `bash_shell` & `zsh_shell`
   - `platform` replaced `cygwin`, `darwin`, `solaris` & `freebsd`
   - `sdkman_channel` replaced `sdkman_beta_channel`
14. standardized & simplified echoes
15. fixed some typos
16. reformatted files: tab indents, spacing, single quotes, double quotes, braces

I can break this PR down in to separate PRs, create issues, modify or remove changes, etc. as you see fit.

I just didn't want to spend time creating tons of separate PRs, creating issues, dealing with separate branches, etc., if unnecessary.

I've thought of additional improvements that I can work on after we've dealt with my current batch of commits.

You should look at the various commits locally on your computer instead of directly on this github PR web page, because the commits seem to be ordered here by when they were first committed, rather than from parent commit to child commit (I rebased a few times to clean up the commits).

Thanks.


marc0der:
Sorry, not going to accept such a _huge_ PR. Please split into multiple small ones, each with a specific intent.

rgoldberg:
I didn't intend fo you to accept the whole PR as one.

I just wanted you to see all the changes so you could indicate the PR granularity that you want.

marc0der:
If you would be happy to split it up, I can review them individually. I'll happily merge them, provided they meet the required standards.

Also, please insure that your commit messages are imperative statements of the intent of the unit of work that you are working on. A message like `sdkman-path-helpers.sh` won't do.

Closing this PR for now.

PR #737, opened 2020-03-27
PR 3: upgrade Gradle to 6.3, upgrade many dependencies, use Gradle Kotlin DSL instead of Groovy DSL

rgoldberg:
Includes commits from PRs 1 & 2.

Upgraded Gradle to 6.3.

Upgraded many dependencies.

Use Gradle Kotlin DSL instead of Groovy DSL.


build.gradle.kts
```diff
@@ -0,0 +1,135 @@
...
+		userOrg  = "sdkman"
+		setLicenses("Apache-2.0")
+		vcsUrl   = "https://github.com/sdkman/sdkman-cli.git"
+		version(closureOf<VersionConfig> {
+			name     = sdkmanCliBuildVersion
+			desc     = "Binary zip distribution of SDKMAN bash client."
+			released = Date().toString()
+		})
+	})
+}
```
marc0der, line 135: This is a groovy project, why are we cutting over to a Kotlin build DSL?

gradlew
```diff
@@ -7,7 +7,7 @@
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#      http://www.apache.org/licenses/LICENSE-2.0
+#      https://www.apache.org/licenses/LICENSE-2.0
```
marc0der, line 10: Not related to upgrading Gradle.

marc0der:
This PR spans 43 files and few of these commits have anything to do with upgrading Gradle to 6.3. I would like to see _only_ the files affected by upgrading Gradle in this PR.

build.gradle.kts
rgoldberg: Kotlin is statically typed.  Groovy isn't.  In IntelliJ, many usages of the Groovy DSL aren't statically verified by the IDE, cannot be navigated to, etc.

gradlew
rgoldberg: This actually is related to upgrading Gradle.  It's the Gradle Wrapper 6.3 version of `gradlew`.  The Gradle team upgraded this script, so I upgraded to match what is distributed with Gradle 6.3.  I didn't look at the change, I just ensured it was the latest version.  Even if the comment change isn't really necessary, it's best to be consistent with the latest Gradle.

rgoldberg:
The other commits should be in PRs 1 & 2.  The PRs are sequential; each should contain the commits of my previous PRs.  We should finish with PR 1 first, then PR 2, etc.  If extra commits are included in this PR other than commits from prior sequential PRs, and other than those specific to this PR, then I will need to compare the branches in github versus those in my local clone, as my local clone is correct.

rgoldberg:
There are conflicts now between my PRs & master, so I'll fetch, merge & rebase to fix them.

Once the conflicts have been resolved:

- start with PR 1: https://github.com/sdkman/sdkman-cli/pull/741
- then do PR 2: https://github.com/sdkman/sdkman-cli/pull/742
- then do this PR
- do all PRs sequentially

If you have any issues with any PR along the way, don't go on to the next one.  Let me fix up the problem PR, then, after you've merged that PR, go on to the next one.

marc0der:
Each PR should be in it's own feature branch, and should be taken directly off master. There should be only commits related to that feature in the branch, and hence in the PR.

rgoldberg:
I've rebased off current `master`, so you should be cleared to look at PR 1 #741.

marc0der:
I made some changes and added them as subsequent commits.
* Went back to the binary distribution of Gradle as the `-all` incurs a huge download penalty.
* I reformatted the build.gradle to have default formatting.
* Went back to the shorthand notation of build dependencies. The `group`, `name`, `version` is a legacy notation and way too verbose.

PR #742, opened 2020-04-02
PR 2: Whitespace formatting

rgoldberg:
Includes commits from PR 1.

Standardized whitespace formatting of files:

- use the same indent style (type & number of characters per indent) for all lines in all files in the same directory
- added & removed some empty lines, mainly in & around conditionals & loops, ate beginning or end or blocks, or at end of file
- made a few if conditions on multiple lines instead of one line
- removed trailing whitespace from whitespace-only lines


marc0der:
Just examined this PR in IDEA and see lots of problems. It seems like you've set your IDE default to spaces, not tabs. This is not default IntelliJ behaviour (they use smart tabbing) and will mess up everyone's setups who do use the default (myself included). I suggest we skip this PR, sorry.

.editorconfig
```diff
@@ -0,0 +1,1324 @@
...
+ij_css_value_alignment = do_not_align
+
+[*.csv]
+max_line_length = 2147483647
+ij_csv_wrap_long_lines = false
+
+[*.feature]
+ij_gherkin_keep_indents_on_empty_lines = false
+
+[*.java]
```
marc0der, line 43: I think this `.editorconfig` can be a lot smaller. I would prefer to see only editor config for the relevant files. we don't really care about `css`, `csv`, `java` etc.

rgoldberg: Done.  Please see force pushed code.

PR #745, opened 2020-04-08
PR 6: Candidate cache update bug

rgoldberg:
In `__sdk_update`, candidates wouldn't be updated if the same number of
candidates were removed & added because detection was solely based on
old vs. new candidate counts.

### DIFF
--- a/src/main/bash/sdkman-update.sh
+++ b/src/main/bash/sdkman-update.sh
@@ -48,26 +48,34 @@ function __sdk_update() {
 			return 0
 		fi
 
-		local fresh_candidates_length=${#fresh_candidates_csv}
-		local cached_candidates_length=${#SDKMAN_CANDIDATES_CSV}
-		__sdkman_echo_debug "Fresh and cached candidate lengths: $fresh_candidates_length $cached_candidates_length"
+		__sdkman_echo_debug "Fresh and cached candidate lengths: ${#fresh_candidates_csv} ${#SDKMAN_CANDIDATES_CSV}"
 
-		local diff_candidates=$(echo ${fresh_candidates[@]} ${cached_candidates[@]} | tr ' ' '\n' | sort | uniq -u | tr '\n' ' ')
-		if (( fresh_candidates_length > cached_candidates_length )); then
-			echo ""
-			__sdkman_echo_green "Adding new candidates(s): $diff_candidates"
-			echo "$fresh_candidates_csv" >| "$SDKMAN_CANDIDATES_CACHE"
-			echo ""
-			__sdkman_echo_yellow "Please open a new terminal now..."
-		elif (( fresh_candidates_length < cached_candidates_length )); then
-			echo ""
-			__sdkman_echo_green "Removing obsolete candidates(s): $diff_candidates"
-			echo "$fresh_candidates_csv" >| "$SDKMAN_CANDIDATES_CACHE"
-			echo ""
-			__sdkman_echo_yellow "Please open a new terminal now..."
+		local combined_candidates diff_candidates
+
+		combined_candidates=("${fresh_candidates[@]}" "${cached_candidates[@]}")
+
+		diff_candidates=($(printf $'%s\n' "${combined_candidates[@]}" | sort | uniq -u))
+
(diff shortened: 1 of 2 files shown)


src/test/cucumber/update.feature
```diff
@@ -18,7 +18,7 @@ Feature: Update
 			| kotlin    |
 			| scala     |
 		When I enter "sdk update"
-		Then I see "Adding new candidates(s): kotlin"
+		Then I see "Setting candidate list to: activator groovy kotlin scala"
```
marc0der, line 21: I do prefer the specific messages stating whether we add/remove candidates, and exactly what was added/removed. In reality this list is _very_ long and will be overwhelming feedback for the user.

src/test/cucumber/update.feature
```diff
@@ -27,7 +27,7 @@ Feature: Update
 			| groovy    |
 			| scala     |
 		When I enter "sdk update"
-		Then I see "Removing obsolete candidates(s): activator"
+		Then I see "Setting candidate list to: groovy scala"
```
marc0der, line 30: As above.

src/test/groovy/sdkman/specs/CandidatesCacheUpdateSpec.groovy
```diff
@@ -97,7 +97,7 @@ class CandidatesCacheUpdateSpec extends SdkmanEnvSpecification {
 		bash.execute("sdk update")
 
 		then:
-		bash.output.contains('Adding new candidates(s): groovy scala')
+		bash.output.contains('Setting candidate list to: groovy scala')
```
marc0der, line 100: As above.

src/main/bash/sdkman-update.sh
```diff
@@ -48,26 +48,19 @@ function __sdk_update() {
...
-			__sdkman_echo_green "Adding new candidates(s): $diff_candidates"
-			echo "$fresh_candidates_csv" > "$SDKMAN_CANDIDATES_CACHE"
-			echo ""
-			__sdkman_echo_yellow "Please open a new terminal now..."
-		elif (( fresh_candidates_length < cached_candidates_length )); then
-			echo ""
-			__sdkman_echo_green "Removing obsolete candidates(s): $diff_candidates"
-			echo "$fresh_candidates_csv" > "$SDKMAN_CANDIDATES_CACHE"
-			echo ""
-			__sdkman_echo_yellow "Please open a new terminal now..."
```
marc0der, line 67: This code is very fragile due to the limitations of bash. It is _very_ rare that we add a candidate, and dropping one is even more rare. I suggest we leave it as it is as it works perfectly, even if a bit naive in that it is based on counts.

marc0der:
The feedback that the user gets trumps the extreme edge case that you are trying to fix here. Doing an update and seeing exactly what was added is very useful information. Adding candidates is rare, removing candidates even rarer, and adding _and_ removing simultaneously is unheard of.

src/main/bash/sdkman-update.sh
```diff
@@ -48,26 +48,34 @@ function __sdk_update() {
...
+				__sdkman_echo_green "\nRemoving obsolete candidates(s): ${delta[*]}"
+			fi
+
+			echo "${fresh_candidates_csv}" >| "${SDKMAN_CANDIDATES_CACHE}"
+			__sdkman_echo_yellow $'\nPlease open a new terminal now...'
 		else
-			touch "$SDKMAN_CANDIDATES_CACHE"
-			__sdkman_echo_green "No new candidates found at this time."
+			touch "${SDKMAN_CANDIDATES_CACHE}"
+			__sdkman_echo_green $'\nNo new candidates found at this time.'
```
marc0der, line 78: Maybe not understanding this, but what is the purpose of the `$` before the string.

rgoldberg: The $ before the string changes backslash escape sequences into what they escape. So `\n` becomes an actual newline character, instead of a backslash followed by an `n`. Many commands will interpret escape sequences, but this removes the necessity to ensure that the command is correctly setup (e.g., this means that you don’t have remember to use `echo -e` instead of just `echo`).

Issue #484, opened 2016-10-18
"file exists" when opening a terminal


cazacugmihai:
When opening the terminal, I have this error:

> /home/$USER/.sdkman/bin/sdkman-init.sh:144: file exists: /home/$USER/.sdkman/var/version

The issue is fixed by running `sdk selfupdate force` but it appears again once the OS is restared.

Environment:
- Ubuntu 16.04
- shell: /bin/zsh
- sdkman: latest version

jverne:
This is because your shell has "noclobber" set, and the sdkman script is trying to clobber a file. You can check this by running a command like this:

    set -o | grep clobber

There isn't a good solution, other than setting the "clobber" shell option for any shell that runs sdkman, prior to running any sdkman script. e.g.:

    set -o clobber

Of course, you can set this in a startup file for all shells, or for a shell that you only run for sdkman.

I suppose you could modify the sdkman scripts so it uses ">|" instead of ">" when overwriting files. If this is done I'm guessing the next update to sdkman will overwrite these changes.

cazacugmihai:
@jverne: I'm using `zprezto` and `noclobber` option is enabled by default. I have disabled it and everything works as expected now. Thank you!

If there is no other solution (implemented in `sdkman`), this issue can be closed.

marc0der:
Closing this issue as the problem lies with prezto being opinionated about file clobbering. I personally use ZSH with ohmyzsh and have no problems.

Worth adding that I also experienced this problem in the past with prezto which ultimately led me to switching to ohmyzsh.

jverne:
This is not a zprezto issue. It is a shell issue. zprezto just also happens to complain when the shell has noclobber set.

Here is what happens on plain bash:

    jverne@YKF-JVERNE:~$ echo foo > out
    jverne@YKF-JVERNE:~$ echo foo > out
    bash: out: cannot overwrite existing file
    jverne@YKF-JVERNE:~$ sdk
    bash: /home/jverne/.sdkman/var/broadcast_id: cannot overwrite existing file
    bash: /home/jverne/.sdkman/var/broadcast: cannot overwrite existing file

jverne:
It occurs to me that a script fix might be to use a subshell to update/clobber files. e.g., in all the places we clobber files, we do something like this:

    ( set + noclobber; do_the_thing_that_clobbers > some_file )

[...]

PR #748, opened 2020-04-16
Don't fail redirections if noclobber is set

majutsushi:
Using the '>|' syntax for shell redirections instead of '>' forces an
overwrite even if the noclobber option is set.

This addresses Issue #484.

I've added the `-o noclobber` option to the bash test environment and confirmed that without my changes some of the tests fail, while with my changes they all pass. This also prevents regressions in case someone adds another redirection later that doesn't use `>|`.

Since I use zsh myself I can confirm that it works there as well.

### DIFF
--- a/src/test/groovy/sdkman/env/BashEnv.groovy
+++ b/src/test/groovy/sdkman/env/BashEnv.groovy
@@ -46,7 +46,7 @@ class BashEnv {
 	 * Starts the external bash process.
 	 */
 	void start() {
-		process = ["bash", "--noprofile", "--norc", "-i"].execute(env, workDir)
+		process = ["bash", "--noprofile", "--norc", "-i", "-o", "noclobber"].execute(env, workDir)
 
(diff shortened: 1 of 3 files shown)


src/test/groovy/sdkman/env/BashEnv.groovy
```diff
@@ -46,7 +46,7 @@ class BashEnv {
 	 * Starts the external bash process.
 	 */
 	void start() {
-		process = ["bash", "--noprofile", "--norc", "-i"].execute(env, workDir)
+		process = ["bash", "--noprofile", "--norc", "-i", "-o", "noclobber"].execute(env, workDir)
```
ironcamel, line 49: This change on this line is not necessary right?

ironcamel: Nevermind, I just ready your explanation in the description. Makes sense 👍

marc0der:
Thanks for the contribution!

PR #749, opened 2020-04-17
Fix cached candidate server api error-handling bug in sdk update

rgoldberg:
Fix cached candidate server api error-handling bug in `__sdk_update`, plus a few other improvements to `__sdk_update`.

### DIFF
--- a/src/main/bash/sdkman-update.sh
+++ b/src/main/bash/sdkman-update.sh
@@ -22,17 +22,13 @@ function __sdk_update() {
 
 	local fresh_candidates_csv=$(__sdkman_secure_curl_with_timeouts "$candidates_uri")
 
-	local fresh_candidates=("")
-	local cached_candidates=("")
-
+	local fresh_candidates
 	if [[ "$zsh_shell" == 'true' ]]; then
-		fresh_candidates=( ${(s:,:)fresh_candidates_csv} )
-		cached_candidates=( ${(s:,:)SDKMAN_CANDIDATES_CSV} )
+		fresh_candidates=(${(s:,:)fresh_candidates_csv})
 	else
 		OLD_IFS="$IFS"
 		IFS=","
 		fresh_candidates=(${fresh_candidates_csv})
-		cached_candidates=(${SDKMAN_CANDIDATES_CSV})
 		IFS="$OLD_IFS"
 	fi
 
@@ -51,7 +47,7 @@ function __sdk_update() {
 
 		local combined_candidates diff_candidates
 
-		combined_candidates=("${fresh_candidates[@]}" "${cached_candidates[@]}")
+		combined_candidates=("${fresh_candidates[@]}" "${SDKMAN_CANDIDATES[@]}")
 
 		diff_candidates=($(printf $'%s\n' "${combined_candidates[@]}" | sort | uniq -u))
 
@@ -64,7 +60,7 @@ function __sdk_update() {
 				__sdkman_echo_green "\nAdding new candidates(s): ${delta[*]}"
 			fi
 
-			delta=("${cached_candidates[@]}" "${diff_candidates[@]}")
+			delta=("${SDKMAN_CANDIDATES[@]}" "${diff_candidates[@]}")
 			delta=($(printf $'%s\n' "${delta[@]}" | sort | uniq -d))
 			if ((${#delta[@]})); then
 				__sdkman_echo_green "\nRemoving obsolete candidates(s): ${delta[*]}"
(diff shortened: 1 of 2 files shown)


rgoldberg:
> This is an old workaround that is required as far as I remember. If you don't set IFS back to what it was, you could leave the user's shell in an inconsistent state.

Doing `<variable>=<value> <some other code on same line>` only sets `<variable>=<value>` for that one command. So, `IFS=',' …` will only apply to that line, and it resets to its previous value after that line is done.

> Why have we dropped the `cached_candidates` local variable here? I'd prefer not to depend on an environment variable in the body of the function. Also, it might be simpler but makes the code less understandable.

The old code already depended on parsing `SDKMAN_CANDIDATES_CSV`.  `SDKMAN_CANDIDATES` is just `SDKMAN_CANDIDATES_CSV` already parsed, so there's no need to parse `SDKMAN_CANDIDATES_CSV` yet again.  `SDKMAN_CANDIDATES` is already used elsewhere in the codebase, used up to 4 times in the same function. `SDKMAN_CANDIDATES` is only used twice in the new code from this PR.  `SDKMAN_CANDIDATES_CACHE` was already used three times, and `SDKMAN_CANDIDATES_CSV` two times, in the same function, so I assumed using `SDKMAN_CANDIDATES` twice was OK.

> It is fine deferring the work, but generally we declare local variables at the top of the function.

I can move variable declarations to the beginning of the function if you want, though I think the assignment should definitely be deferred to where I moved it.  I just generally try to have code avoid performing unnecessary work, not just for performance, but also so that developers reading code don't wonder what code is for many lines before it's actually useful.

> We would need a test to prove that this is a bug and that this fixes it.

I think this bug is easily observable from the code. In the existing code,`fresh_candidates` was was grepped before it was ever declared or assigned, and it's later assigned to an array, not a scalar.  The original author obviously meant to grep `fresh_candidates_csv` instead.  I fixed that bug, avoided the unnecessary `tr` call by using `grep -i`, and delayed the grep until it was actually needed.

src/main/bash/sdkman-update.sh
```diff
@@ -50,9 +35,15 @@ function __sdk_update() {
 
 		__sdkman_echo_debug "Fresh and cached candidate lengths: ${#fresh_candidates_csv} ${#SDKMAN_CANDIDATES_CSV}"
 
-		local combined_candidates diff_candidates
+		local fresh_candidates combined_candidates diff_candidates
+
+		if [[ "${zsh_shell}" == 'true' ]]; then
+			fresh_candidates=(${(s:,:)fresh_candidates_csv})
+		else
+			IFS=',' read -a fresh_candidates <<< "${fresh_candidates_csv}"
```
marc0der, line 43: This is broken in ZSH: `read: bad option: -a`

src/main/bash/sdkman-update.sh
```diff
@@ -50,9 +35,15 @@ function __sdk_update() {
...
+		fi
 
-		combined_candidates=("${fresh_candidates[@]}" "${cached_candidates[@]}")
+		combined_candidates=("${fresh_candidates[@]}" "${SDKMAN_CANDIDATES[@]}")
```
marc0der, line 46: Despite the counter arguments, I insist on using local variables please. If there are other occurances of this, they should be fixed too. Global variable access is never a good thing, especially if such variables are mutable.

marc0der:
Regarding the writing of a test: apologies in advance but if you are fixing a bug, it should always go with a test to prove it.

src/main/bash/sdkman-update.sh
marc0der: Fair points, it would be nice to assign it to a local variable early though if we can't pass it through as a param.

marc0der:
I ran the following on the command line:
```
$ IFS=',' read -a fresh_candidates <<< "1,2,3,4,5"
read: bad option: -a

$ zsh --version
zsh 5.8 (x86_64-pc-linux-gnu)
```

It works fine in bash, not in ZSH.

rgoldberg:
Not sure if you saw my comment elsewhere, but that `read -a` line won’t ever run in zsh, because of the if condition.

src/main/bash/sdkman-update.sh
marc0der: Never mind, I see the if statement above ;-)

marc0der:
Thanks for your patience @rgoldberg

rgoldberg:
Thanks.  I understand.  You're busy, and I need to learn all your preferences for sdkman PRs.  Will submit the next PR.

rgoldberg:
Gotta rebase all the `sdk env` changes first.  Once done, will submit PR

Issue #603, opened 2017-09-06
Provide a way to manage aliases for a major version of a tool


vpavic:
As discussed in the [sdkman/user-issues](https://gitter.im/sdkman/user-issues?at=59ae7cf3ee5c9a4c5f352491), it would be nice to have SDKMAN provide a way to manage additional aliases which would point to the latest major version of a tool.

This would be of great help when working with IDEs, since we would have a well defined aliases of, for example, major JDK releases: JDK 8, JDK 7, etc. So when a new update of a given JDK release would be installed, the alias would be updated to point to it and therefore no change to IDE configuration would be required.

I'm currently managing something like this by manually creating symlinks:

```shell
$ ll .sdkman/candidates/java/
total 28
drwxr-xr-x  7 vedran vedran 4096 Aug 31 07:59 ./
drwxr-xr-x 10 vedran vedran 4096 Jul  6 09:31 ../
drwxr-xr-x  9 vedran vedran 4096 Apr 18 17:40 6u93-zulu/
drwxr-xr-x  9 vedran vedran 4096 Apr 17 15:41 7u141-zulu/
drwxr-xr-x  8 vedran vedran 4096 Jul 12 13:41 8u141-oracle/
drwxr-xr-x  9 vedran vedran 4096 Jul 26 16:26 8u144-zulu/
drwxr-xr-x 10 vedran vedran 4096 Aug 21 16:59 9ea14-zulu/
lrwxrwxrwx  1 vedran vedran   47 Aug  9 09:09 current -> /home/vedran/.sdkman/candidates/java/8u144-zulu/
lrwxrwxrwx  1 vedran vedran    9 Aug 22 14:04 java-6-zulu -> 6u93-zulu/
lrwxrwxrwx  1 vedran vedran   10 Aug 22 14:04 java-7-zulu -> 7u141-zulu/
lrwxrwxrwx  1 vedran vedran   12 Aug 31 07:59 java-8-oracle -> 8u141-oracle/
lrwxrwxrwx  1 vedran vedran   10 Aug 22 14:04 java-8-zulu -> 8u144-zulu/
lrwxrwxrwx  1 vedran vedran   10 Aug 27 20:13 java-9-zulu -> 9ea14-zulu/
```

Which results in:

```shell
$ sdk ls java

================================================================================
Available Java Versions
================================================================================
   + java-9-zulu                                                                   
   + java-8-zulu                                                                   
   + java-8-oracle                                                                 
[...]

PR #750, opened 2020-04-18
Link major version on install

artgon:
This addresses Issue #603

I think this should work for other SDKs, but I'm particularly interested in Java versions for IDE support.

The limitation of this PR is that it will override the link without confirmation, but should work in general cases of SDK upgrades.


marc0der:
We most certainly require tests for a new feature to be considered.

Another problem that I see is that it simply creates a link pointing to what was last installed. It doesn't track the latest version of a specific major release. Any ideas how we could solve this?

marc0der:
No movement on this for well over a year so closing.

PR #751, opened 2020-04-19
Add command to generate `.sdkmanrc` file

helpermethod:

This addresses Issue #730


helpermethod:
Hi @marc0der,

this is the current WIP of the `sdk env init` command.

Some open questions:

* Which Java version should I add as an example? Is there some easy way to get the currently used Java version, and fall back on some default?
* If `sdk env` is called with some "subcommand" other than `init` what would be the expected outcome? Should an error be displayed? Should just `sdk env` be called?

src/main/bash/sdkman-env.sh
```diff
@@ -30,14 +30,26 @@ function __sdk_env() {
 	local line_number=0
 
 	while IFS= read -r line || [[ -n $line ]]; do
-		if [[ ! $line =~ ^[[:lower:]]+=.+$ ]]; then
-			__sdkman_echo_red "${sdkmanrc}:${line_number}: Invalid candidate format! Expected 'candidate=version' but found '$line'"
+		__sdkman_is_blank_or_comment "$line" && continue
+
+		if ! __sdkman_matches_candidate_format "$line"; then
+			__sdkman_echo_red "${sdkrc}:${line_number}: Invalid candidate format! Expected 'candidate version' but found '$line'"
```
marc0der: Can we split this into two lines please, separated by a newline .
The warning in red, the further information in yellow.

src/main/bash/sdkman-env.sh
```diff
@@ -30,14 +30,26 @@ function __sdk_env() {
...
+		local candidate version
+		IFS=' \t' read -r candidate version <<< "$line"
+		__sdk_use "$candidate" "$version"
 
 		((line_number++))
-	done < "$sdkmanrc"
+	done < "$sdkrc"
+}
+
+function __sdkman_is_blank_or_comment() {
```
marc0der: how about `__sdkman_line_is_blank_or_comment()`?

src/main/bash/sdkman-env.sh
```diff
@@ -30,14 +30,26 @@ function __sdk_env() {
...
+	[[ $1 =~ ^[[:blank:]]*|\#.*$ ]]
 }
+
+function __sdkman_matches_candidate_format() {
```
marc0der, line 58: `__sdkman_line_matches_candidate_format()`?

src/main/bash/sdkman-env.sh
```diff
@@ -30,26 +30,26 @@ function __sdk_env() {
...
+		IFS=' \t' read -r candidate version rest <<< "$line"
 		__sdk_use "$candidate" "$version"
 
 		((line_number++))
 	done < "$sdkrc"
 }
 
-function __sdkman_is_blank_or_comment() {
-	[[ $1 =~ ^[[:blank:]]*|\#.*$ ]]
+function __sdkman_is_blank_line_or_comment_line() {
```
marc0der: Lol, don't worry. I see it got addressed in a later commit.

src/main/bash/sdkman-env.sh
```diff
@@ -17,10 +17,10 @@
 #
 
 function __sdk_env() {
-	readonly sdkrc='.sdkrc'
+	readonly sdkmanrc='.sdkmanrc'
```
marc0der, line 20: I like this better ;-)

src/main/bash/sdkman-env.sh
```diff
@@ -19,30 +19,35 @@
...
+		cat <<- EOF > "$sdkmanrc"
+		# This file was created by SDKMAN.
+		# 
+		# java=8.0.252-zulu 
+		EOF
+	fi
+
 	if [[ ! -f "$sdkmanrc" ]]; then
-		__sdkman_echo_red "No $sdkmanrc file found."
+		__sdkman_echo_red "SDKMAN can't find an .sdkmanrc file in your current directory."
```
marc0der, line 31: How about _Could not find .sdkmanrc file in current directory._?

src/main/bash/sdkman-env.sh
```diff
@@ -19,30 +19,35 @@
...
 		echo ""
-		__sdkman_echo_yellow "Please create one before using this command."
+		__sdkman_echo_yellow "We recommend creating one by entering 'sdk env init'."
```
marc0der, line 33: _Run 'sdk env init' to create it._

marc0der:
Apologies, I was reviewing this commit by commit so not all the comments make sense by the end.

helpermethod:
@marc0der Thanks for the valuable suggestions, I suck at error messages / warnings ;).

I've also simplified things a bit, the lines are now normalised (stripped by whitespace and comments) before performing the format check, which will help the user in case he entered something invalid.

I've also fallen back to using `=` for separating candidate and version, as some devs may already use the beta version (and it actually doesn't make a difference what delimiter is used).

marc0der:
Sounds sensible yes. Haven't looked into it, but the build on your branch seems to be failing. Once we have a green build I'd be happy to merge.

helpermethod:
Hi @marc0der,

I've fixed the outstanding issues, added some more tests, fixed some broken tests and changed how the `.sdkmanrc` file is created (it inserts the currently used java version or falls back to some fixed java version).

There are still some untested code branches and minor things which can be improved but IMHO nothing which prevents a release.

helpermethod:
It seems I can use `version=​$(​__sdkman_secure_curl ​"​${SDKMAN_CANDIDATES_API}​/candidates/default/​${candidate}​"​)​` to determine the fallback Java version?

src/test/groovy/sdkman/specs/EnvCommandSpec.groovy
```diff
@@ -24,7 +31,7 @@ class EnvCommandSpec extends SdkmanEnvSpecification {
 			}
 		}
 
-		new File(bash.workDir, '.sdkmanrc').text = sdkmanrc
+		new File(bash.workDir, '.sdkmanrc').text = sdkrc
```
marc0der, line 34: Should this be `sdkmanrc`?

src/test/groovy/sdkman/specs/EnvCommandSpec.groovy
```diff
@@ -36,20 +43,50 @@ class EnvCommandSpec extends SdkmanEnvSpecification {
 		}
 
 		where:
-		sdkmanrc << ["grails=2.1.0\ngroovy=2.4.1", "grails=2.1.0\ngroovy=2.4.1\n"]
+		sdkrc << [
```
marc0der, line 46: Same as above.

src/test/groovy/sdkman/specs/EnvCommandSpec.groovy
```diff
@@ -36,20 +43,50 @@ class EnvCommandSpec extends SdkmanEnvSpecification {
...
+		new File(bash.workDir, ".sdkmanrc").text = sdkrc
+
+		when:
+		bash.execute("sdk env")
+
+		then:
+		bash.output.contains("Using groovy version 2.4.1 in this shell.")
+
+		where:
+		sdkrc << [
```
marc0der, line 86: It is just naming, but let's call it `sdkmanrc` throughout.

marc0der:
Some very minor naming bits  in the tests, but almost ready to merge! Also, I hope you don't mind me squashing your commits.

src/test/groovy/sdkman/specs/EnvCommandSpec.groovy
helpermethod: Consistency is important, I agree.

helpermethod:
Hi @marc0der,

I've added one more test and performed the suggested fixes.

src/main/bash/sdkman-env.sh
```diff
@@ -17,27 +17,67 @@
 #
 
 function __sdk_env() {
-	readonly sdkmanrc='.sdkmanrc'
+	SDKMANRC=".sdkmanrc"
```
marc0der, line 20: Was there a problem with `readonly`? I would even be happy with it being `local`, but don't really like it defined as global though.

src/main/bash/sdkman-env.sh
```diff
@@ -17,27 +17,67 @@
...
 
-	if [[ ! -f "$sdkmanrc" ]]; then
-		__sdkman_echo_red "No $sdkmanrc file found."
+	if [[ "$1" == "init" ]]; then
```
marc0der, line 22: What exactly is `"$1"` here? I would suggest assigning whatever it is to a local variable and then using that here instead.

src/main/bash/sdkman-env.sh
```diff
@@ -17,27 +17,67 @@
...
+
+	echo "java=$version" > "$SDKMANRC"
+
+	__sdkman_echo_green "$SDKMANRC created."
+}
 
-		((line_number++))
-	done < "$sdkmanrc"
+function __sdkman_is_blank_line() {
+	[[ -z "$1" ]]
```
marc0der, line 72: What will this return? An exit code?

src/main/bash/sdkman-env.sh
rgoldberg: I don't know why @helpermethod mad it a global.  `readonly` variables fail in bash 3 when a method is called more than once, as `readonly` in bash 3 creates a readonly global.  The way around this is to use `local -r` which works in bash 3, as well as in bash 5 & in zsh.  If that's why he changed it, using `local -r` should fix the problem.

helpermethod: I will make it local and pass it explicitly to the _generate function.

src/main/bash/sdkman-env.sh
helpermethod: From the bash manpage

> When executed, the exit status of a function is the exit status of the last command executed in the body.

src/main/bash/sdkman-env.sh
helpermethod: I've mad the variable local and readonly (local - r).

src/main/bash/sdkman-env.sh
helpermethod: Assigned $1 to a variable.

PR #754, opened 2020-04-28
Simplify bash csv parsing in sdkman-init.sh

rgoldberg:
Simplify bash csv parsing in `sdkman-init.sh` by changing `IFS` just for one command, like what was done recently in `sdkman-update.sh`.

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -82,7 +82,7 @@ for f in "${scripts[@]}"; do
 	source "$f"
 done
 IFS="$OLD_IFS"
-unset scripts f
+unset OLD_IFS scripts f
 
 # Load the sdkman config if it exists.
 if [ -f "${SDKMAN_DIR}/etc/config" ]; then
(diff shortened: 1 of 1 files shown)


src/main/bash/sdkman-init.sh
```diff
@@ -82,7 +82,7 @@ for f in "${scripts[@]}"; do
 	source "$f"
 done
 IFS="$OLD_IFS"
-unset scripts f
+unset OLD_IFS scripts f
```
marc0der, line 85: Why unset `OLD_IFS` at all if it is no longer being used?

rgoldberg: `OLD_IFS` is still used. It used to be used in 2 places but is now just used in 1.

marc0der: Is it worth cutting over to the new way in the other place too for consistency? (Only if this isn't a huge thing to do, of course). Then we don't have to worry about managing this nasty global state in the shell.

rgoldberg: I would have if this syntax would work where it is still used, but it won't.

The place where `IFS=$'\n'` is needed is:

```bash
scripts=($(find "${SDKMAN_DIR}/src" "${SDKMAN_DIR}/ext" -type f -name 'sdkman-*'))
```

You can't use `<variable>=<value> ` in the same command a permanent assignment.

The shells just don't allow it.

Maybe it could work inside a command substitution.

I'm actually trying to get a glob replacement for the `find` command to work.  The glob wouldn't use `IFS` at all, and wouldn't use an external command like `find`.  It seems to work from the command line, but not in the Gradle tests.  I imagine it has something to do with the bash environment settings, but I don't know what yet.

I'll add a PR for that once I get it working (and I'll change as many `find` usages to globs as I can easily do).

rgoldberg: Looks like it works inside the command substitution.  Don't accept this PR yet.  Will push an update in a few minutes.

rgoldberg:
I'm not sure if I should have resolved the previous conversation, because I don't know the preferred github process, but wouldn't allow me to add any further comments in that conversation.

I've gotten globs (which don't use `IFS`) working in a later commit, so there shouldn't be a need to cleanup more `IFS` uses in this PR.  We shouldn't move the changes from `find` to globs here because of rebasing / merging issues, and there are a few things to discuss about globs before we merge their usages in.  It doesn't make sense to discuss those issues until we get to them, however.

PR #758, opened 2020-05-02
Improve __sdk_env

rgoldberg:
- Don't restrict candidate names to lowercase letters (only require that candidate name doesn't contain `=`).
 - Fail immediately when there's an invalid candidate & version pair, since invalid line format also fails immediately
 - Parse `candidate` until first `=`, not until last `=`, but still parse `version` from first `=` (fixes problem when more than one `=` on a line).
 - Make `line` a local variable
 - Move `normalised_line` declaration outside of `while` to avoid message on iterations after the first
 - Improve error message
 - Remove unnecessary slash from ends of search & replaces in `__sdkman_normalise`

### DIFF
--- a/src/main/bash/sdkman-env.sh
+++ b/src/main/bash/sdkman-env.sh
@@ -34,21 +34,22 @@ function __sdk_env() {
 		return 1
 	fi
 
-	local normalised_line
+	local line normalised_line
 	while IFS= read -r line || [[ -n "$line" ]]; do
 		normalised_line="$(__sdkman_normalise "$line")"
 
 		__sdkman_is_blank_line "$normalised_line" && continue
 
 		if ! __sdkman_matches_candidate_format "$normalised_line"; then
-			__sdkman_echo_red "Invalid candidate format!"
-			echo ""
-			__sdkman_echo_yellow "Expected 'candidate=version' but found '$normalised_line'"
+			__sdkman_echo_red $'\nInvalid candidate format!'
+			__sdkman_echo_yellow "\nExpected '<candidate>=<version>' but found '${normalised_line}'"
 
 			return 1
 		fi
 
-		__sdk_use "${normalised_line%=*}" "${normalised_line#*=}"
+		if ! __sdk_use "${normalised_line%%=*}" "${normalised_line#*=}"; then
+			return 1
+		fi
 	done < "$sdkmanrc"
 }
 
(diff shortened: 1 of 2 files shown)


helpermethod:
Some nice improvements, thanks!

marc0der:
Looks good, would be good to have some tests to assert some of these improvements though.

helpermethod:
Most of these are refactorings and already covered by tests. Though it wouldn't hurt to add 1 or 2 additional cases for the changes to the validation and splitting logic.

rgoldberg:
Not to be contrarian, but every test slows down every build, and adds to the scenarios that a dev might have to read.

If the minor bugs hadn't been there, but were instead avoided as per the code in this PR, you wouldn't have requested tests for that functionality (if you would have requested tests for that functionality, you would have requested the tests for the existing `env` code, and the minor bugs would have been found and prevented by those tests).

It seems overkill to test all these minor edge cases, like more than one `=` in a line, not restricting candidates to lowercase letters, not outputting extraneous confusing messages, etc.

For the last one, the code could output other messages besides the one we've seen so far, so the best way to ensure that there is no extra output is to change every test from checking if the output contains a string to checking if the output equals a string.  But that should probably be a change to all tests, not just the ones for `env`.

rgoldberg:
I modified an existing test to additionally cover more than one `=` on a line, and non-lowercase characters in candidate names.  I'll push it in a minute or so.

rgoldberg:
Pushed the modifications to env tests.  Covers the changes to `.sdkmanrc` line parsing.  Doesn't cover not outputting extra messages, because code could output any extra messages, not just the currently observed one.  Also, I think that might only occur on zsh, which isn't tested by the tests.

marc0der:
Right, now that you have demonstrated your intention through a test, I (and everyone else) understand what it is that you are trying to achive.

Here are some observations:
* candidates are _always_ expressed in lower case, with no caps or special characters allowed. this is not enforced anywhere, but is a convention throughout SDKMAN.
* we should only accept simple key value pairs, separated by a singe `=`. Multiple `=` on a line should simply not be allowed.

In the end, the simplest approach trumps all. Why accommodate such odd edge cases when they should never occur in the real world? When/if these things become a problem, we will address them at that point.

rgoldberg:
I accommodated edge cases because doing so didn't add any complexity, and it designs for extensibility.

e.g., if there's an sdk called `plan9` that someone wants on sdkman, the existing `sdk env` will fail on that candidate name.  Would you force someone to name the candidate plannine?  That doesn't make sense to me.

If, for some reason, there's a version that includes an `=`, then the existing code will fail on it.

It's one thing thing if I complicated logic to support this, but my change adds no complexity at all, and it reduces complexity of the system by removing artificial constraints, and, more importantly, removes constraints from places where they shouldn't be enforced.  If you want to enforce that all candidate names follow a certain pattern, `__sdkman_env` is not the place to do it.

helpermethod:
Maybe we can rollback point 1 and 3 and keep the other fixes (candidate name and candidate parsing)? If these really become a problem we can revisit them. Just so we can get the feature out 🙂.

marc0der:
I've made the single change we needed (logging issue in ZSH) myself for now so as to unblock the release of your functionality @helpermethod.

PR #763, opened 2020-05-24
Improve __sdkman_secure_curl_download

rgoldberg:
Sequence curl arguments using array instead of scalar in `__sdkman_secure_curl_download`:

- Code is now shell-agnostic
- Fixes argument concatenation bugs


marc0der:
Thanks for the fix!

PR #764, opened 2020-05-25
Remove legacy bash workaround from __sdk_update

rgoldberg:
Remove legacy bash workaround from `__sdk_update`.

The code that didn't work on legacy bash was removed a long time ago, so the workaround is no longer necessary.

The workaround prevents 4 or so tests from completing successfully on legacy bash, so it's very useful to remove the workaround, so we can run the test suite on bash 3.2.x.


marc0der:
Thanks!

Issue #640, opened 2018-04-26
SDKMAN is able to install on Java on ARM


scuba10steve:
We require a conversation to take place in the appropriate Gitter Room before raising a new Github Issue. Please note that **issues will be closed immediately if prior discussion did not take place**. We need to do this to help manage the quality and validity of Issues raised on this project.

Please tick one:
- [ ] this is a new Feature Request, a conversation was started in the [Gitter CLI Development Room](https://gitter.im/sdkman/cli-dev).
- [x] this is a new Bug Report or Issue, a conversation was started in [Gitter User Issues Room](https://gitter.im/sdkman/user-issues).

Please explain the Issue / Feature Request here:

Currently, I am able to use sdkman and install java for the wrong architecture on a raspberry pi.  It looks like it isn't doing some detection for the raspberry pi to prevent the install and is downloading an x86 64 bit version of java.

See error below:
```
pi@raspberrypi:~ $ sdk install java

Downloading: java 8.0.163-zulu

In progress...

######################################################################## 100.0%

Repackaging Java 8.0.163-zulu...

Done repackaging...

Installing: java 8.0.163-zulu
Done installing!

Do you want java 8.0.163-zulu to be set as default? (Y/n): Y

Setting java 8.0.163-zulu as default.
pi@raspberrypi:~ $ java -version
-bash: /home/pi/.sdkman/candidates/java/current/bin/java: cannot execute binary file: Exec format error
pi@raspberrypi:~ $ which java
/home/pi/.sdkman/candidates/java/current/bin/java
pi@raspberrypi:~ $ file
Usage: file [-bcEhikLlNnprsvzZ0] [--apple] [--extension] [--mime-encoding] [--mime-type]
            [-e testname] [-F separator] [-f namefile] [-m magicfiles] file ...
       file -C [-m magicfiles]
       file [--help]
pi@raspberrypi:~ $ file /home/pi/.sdkman/candidates/java/current/bin/java
/home/pi/.sdkman/candidates/java/current/bin/java: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2, for GNU/Linux 2.6.18, BuildID[sha1]=f1ac5b47218b2ffcda98cb143d7f9fc431f3574e, not stripped
```

specific version of sdkman cli: 'SDKMAN 5.6.3+299'

marc0der:
[...]

PR #769, opened 2020-06-28
Add support for Linux ARM64

dvdkruk:
Connected to issue: #640

This adds support for platform detection of Linux ARM64 (AArch64).

I had to change code in the following places to make this works:
- [x] sdkman/sdkman-db-migrations#276
- [x] sdkman/sdkman-hooks#18
- [x] sdkman/sdkman-candidates#13
- [x] sdkman/sdkman-broker#2
- [x] sdkman/vendor-release#1

Hopefully I got all the places covered ;-)

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -34,6 +34,8 @@ SDKMAN_PLATFORM="$(uname)"
 if [[ "$SDKMAN_PLATFORM" == 'Linux' ]]; then
 	if [[ "$(uname -m)" == 'i686' ]]; then
 		SDKMAN_PLATFORM+='32'
+	elif [[ "$(uname -m)" == 'aarch64' ]]; then
+		SDKMAN_PLATFORM+='ARM64'
 	else
 		SDKMAN_PLATFORM+='64'
 	fi
--- a/src/test/groovy/sdkman/support/UnixUtils.groovy
+++ b/src/test/groovy/sdkman/support/UnixUtils.groovy
@@ -3,7 +3,14 @@ package sdkman.support
 class UnixUtils {
 
 	static getPlatform() {
-		asSdkmanPlatform(System.getProperty("os.name"))
+		def os = System.getProperty("os.name")
+		def arch = System.getProperty("os.arch")
+
+		if("aarch64".equals(arch)) {
+			asSdkmanPlatform("LinuxARM64")
+		} else {
+			asSdkmanPlatform(os)
+		}
 	}
 
 	static asSdkmanPlatform(platform) {
--- a/src/test/groovy/sdkman/support/UnixUtils.groovy
+++ b/src/test/groovy/sdkman/support/UnixUtils.groovy
@@ -7,7 +7,7 @@ class UnixUtils {
 	}
 
 	static asSdkmanPlatform(platform, architecture = null) {
-		if("aarch64".equals(architecture)) {
+		if("aarch64" == architecture) {
 			platform += architecture
 		}
 
@@ -27,7 +27,7 @@ class UnixUtils {
 				break
 			case "Linuxaarch64":
 				result = "LinuxARM64"
-				break;
+				break
 			case "FreeBSD":
 				result = "FreeBSD"
 				break
--- a/src/test/groovy/sdkman/support/UnixUtils.groovy
+++ b/src/test/groovy/sdkman/support/UnixUtils.groovy
@@ -3,17 +3,14 @@ package sdkman.support
 class UnixUtils {
 
 	static getPlatform() {
-		def os = System.getProperty("os.name")
-		def arch = System.getProperty("os.arch")
-
-		if("aarch64".equals(arch)) {
-			asSdkmanPlatform("LinuxARM64")
-		} else {
-			asSdkmanPlatform(os)
-		}
+		asSdkmanPlatform(System.getProperty("os.name"), System.getProperty("os.arch"))
 	}
 
-	static asSdkmanPlatform(platform) {
+	static asSdkmanPlatform(platform, architecture = null) {
+		if("aarch64".equals(architecture)) {
+			platform += architecture
+		}
+
 		def result
 		switch (platform) {
 			case "Mac OS X":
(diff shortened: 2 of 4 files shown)


marc0der:
I like where this is going, but not certain about the implementation.

It feels like we're muddying the waters between the OS name and architecture. Since these are two distinct concepts, I think we should represent them as such in the code. wdyt?

src/main/bash/sdkman-init.sh
```diff
@@ -34,6 +34,8 @@ SDKMAN_PLATFORM="$(uname)"
 if [[ "$SDKMAN_PLATFORM" == 'Linux' ]]; then
 	if [[ "$(uname -m)" == 'i686' ]]; then
 		SDKMAN_PLATFORM+='32'
+	elif [[ "$(uname -m)" == 'aarch64' ]]; then
```
marc0der, line 37: Can we expand this to include 32-bit ARM too?

src/test/groovy/sdkman/support/UnixUtils.groovy
```diff
@@ -3,10 +3,14 @@ package sdkman.support
 class UnixUtils {
 
 	static getPlatform() {
-		asSdkmanPlatform(System.getProperty("os.name"))
+		asSdkmanPlatform(System.getProperty("os.name"), System.getProperty("os.arch"))
 	}
 
-	static asSdkmanPlatform(platform) {
+	static asSdkmanPlatform(platform, architecture = null) {
```
marc0der, line 9: Having a `null` default doesn't feel right. You could just leave it off as it will always be set.

src/main/bash/sdkman-init.sh
dvdkruk: I believe it is `armv7l`. However, it is a bit more complicated with ARM32 as there is another dimension of architecture that support hardware floating points (hf) vs software floating points (sf). I don't currently have a system to test this, so didn't want go in this. ARM64 (AArch64) is a little bit easier as hf included in that architecture.

src/test/groovy/sdkman/support/UnixUtils.groovy
dvdkruk: This method is also used in the cucumber steps with only 1 parameter. Without the default value, we either need to create another overloaded method with the default architecture or change all the places where this method is used to provide the 2 parameters. Thought this was the easiest way (least amount of code) to solve it now. What do you think?

dvdkruk:
> I like where this is going, but not certain about the implementation.
> 
> It feels like we're muddying the waters between the OS name and architecture. Since these are two distinct concepts, I think we should represent them as such in the code. wdyt?

Yeah, it is a bit muddled with the different OSs and architectures. With Windows and Apple now focusing on ARM as well, it might be beneficial to restructure the code to reflect that as well.

src/main/bash/sdkman-init.sh
dvdkruk: Just thinking out loud, we can also concatenate the OS and architecture here and send that to the server. This way we don't need to update the client and supporting a new architecture can be by adding support for it on the server side.

marc0der: Not completely sure if I understand, would you be able to give an example of what you mean?

src/test/groovy/sdkman/support/UnixUtils.groovy
marc0der: I'd prefer to always pass it in explicitly if possible, even if it means updating the cucumber features.

marc0der:
I'm happy for this to go in, provided the `asSdkmanPlatform()` method in the tests is changed.

We can look at refining the `sdkman-init.sh` at a later date.

src/test/groovy/sdkman/support/UnixUtils.groovy
dvdkruk: No problem, I have change them in the steps

src/main/bash/sdkman-init.sh
dvdkruk: Instead of converting the platform here we can juts return `"$(uname)_$(uname -m)"` here and provide that to the server. For example,
- `Linux_x86_64` instead of `Linux64`
- `Linux_i686` instead of `Linux32`
- `Linux_aarch64` instead of `LinuxARM64`

But is see the will tightly couple it to `uname` which might be less desirable.

marc0der: Okay, got it. The problem with this is that we need to remain backward compatible. This would result in some nasty business on the server side.

That said, we could possibly do it but I would prefer if we handle this completely on the client side.

pnavais:
Anxiously waiting for this PR to happen ! Would this mean that ARM64 JDKs can be targeted as per [Amazon Corretto's one](https://docs.aws.amazon.com/corretto/latest/corretto-8-ug/downloads-list.html) ?

src/main/bash/sdkman-init.sh
dvdkruk: We can leave this for a later discussion, I think this one is good to go for now.

dvdkruk:
@marc0der I believe this one is good to go now, what do think?

dvdkruk:
@pnavais it should be possible when Amazon Corretto AArch64 is added to the https://github.com/sdkman/sdkman-db-migrations

dvdkruk:
image
Just tested this build and installed the different java version and everything looks good 🥳

marc0der:
@dvdkruk Fantastic news, I'll release it shortly in the stable channel. Thank you for all your hard work to make this happen. Great job!

dvdkruk:
Merged in commits: 134d506e14031ba9dbb6ca6ba36a028ed86cadb1, cf436ed8f26ad3575ba44d4ecd2fe8bcbe90ab0f, eadbbe08401f03c382358232bcb1664830a7ce57, 443930d66d7f26aae74c37c15b5cb5c6c791ff96 & 3bae641050f54eac6ca97cbf612e798e98383827

And released with version 5.9.0 of SKDMAN! 🥳 Enjoy the Linux ARM64 support!

PR #775, opened 2020-07-15
Replace $(cat file) invocations with $(< file) for better performance

### LINKED ISSUE #774: Feature: Replace invocations of $(cat file) with $(< file)

Because SDKMAN takes aliases into account, it may happen that users have aliased `cat` to its popular replacement [bat](https://github.com/sharkdp/bat).

While `bat` is functionally equivalent to `cat`, it is also slower than `cat` by a measurable margin. This especially poses a problem in the context of `sdk env`, where the use of `bat` adds a whooping 300ms to its execution (about 1/3 of its total execution time).

This PR replaces every every invocation of `$(cat file)` in the context of the `sdk env` command by the equivalent substitution `$(< file)` (a special form of command substitution supported by Bash and Zsh).

In the case that users don't have that alias set, the change still improves `sdk env`'s performance, though only by a very small margin.

helpermethod:
Fixes #774 .

### DIFF
--- a/src/main/bash/sdkman-availability.sh
+++ b/src/main/bash/sdkman-availability.sh
@@ -75,11 +75,11 @@ function __sdkman_update_broadcast() {
 	broadcast_old_id=""
 
 	if [[ -f "$broadcast_id_file" ]]; then
-		broadcast_old_id=$(cat "$broadcast_id_file")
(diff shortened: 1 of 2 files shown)


marc0der:
Thank you!

PR #777, opened 2020-07-15
Do not return error when package is already installed

### LINKED ISSUE #731: sdk install should be no-op if package already installed
Right now this is an error, e.g.

    Stop! java 13.0.2.hs-adpt is already installed.

This makes it awkward for scripting

felipecrs:
Fixes #731.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,4 @@ mongo.json
 out/
 sdkman-cli.iml
 target/
+bin/test/
--- a/src/test/cucumber/install_candidate.feature
+++ b/src/test/cucumber/install_candidate.feature
@@ -35,8 +35,8 @@ Feature: Install Candidate
 		And the candidate "grails" version "1.3.9" is available for download
 		And the candidate "grails" version "1.3.9" is already installed and default
 		When I enter "sdk install grails 1.3.9"
-		Then I see "Stop! grails 1.3.9 is already installed."
-		And the exit code is 1
+		Then I see "grails 1.3.9 is already installed."
+		And the exit code is 0
 
 	Scenario: Install a candidate and auto-answer to make it default
 		Given the system is bootstrapped
--- a/src/test/cucumber/service_unavailable.feature
+++ b/src/test/cucumber/service_unavailable.feature
@@ -68,7 +68,8 @@ Feature: Service Unavailable
 		Given the candidate "grails" version "2.1.0" is already installed and default
 		And the system is bootstrapped
 		When I enter "sdk install grails 2.1.0"
-		Then I see "Stop! grails 2.1.0 is already installed."
+		Then I see "grails 2.1.0 is already installed."
+		And the exit code is 0
 
 	# uninstall command
 
--- a/src/main/bash/sdkman-install.sh
+++ b/src/main/bash/sdkman-install.sh
@@ -28,8 +28,8 @@ function __sdk_install() {
 
 	if [[ -d "${SDKMAN_CANDIDATES_DIR}/${candidate}/${VERSION}" || -L "${SDKMAN_CANDIDATES_DIR}/${candidate}/${VERSION}" ]]; then
 		echo ""
-		__sdkman_echo_red "Stop! ${candidate} ${VERSION} is already installed."
-		return 1
+		__sdkman_echo_yellow "${candidate} ${VERSION} is already installed."
+		return 0
 	fi
 
 	if [[ ${VERSION_VALID} == 'valid' ]]; then
--- a/src/test/cucumber/local_developement_versions.feature
+++ b/src/test/cucumber/local_developement_versions.feature
@@ -17,7 +17,8 @@ Feature: Local Development Versions
 		And the candidate "groovy" version "2.1-SNAPSHOT" is already linked to "/tmp/groovy-core"
 		And the system is bootstrapped
 		When I enter "sdk install groovy 2.1-SNAPSHOT /tmp/groovy-core"
-		Then I see "Stop! groovy 2.1-SNAPSHOT is already installed."
+		Then I see "groovy 2.1-SNAPSHOT is already installed."
+		And the exit code is 0
 		And the candidate "groovy" version "2.1-SNAPSHOT" is linked to "/tmp/groovy-core"
 
 	Scenario: Uninstall a local development version


felipecassiors:
Should I improve the output message by prepending "The package" on it?

marc0der:
Looks good so far. Could I please ask you to also add a step in the feature that checks that the return code is 0?

felipecassiors:
For some reason, I don't know if it is because of VS Code or Gitpod itself, every time I clone the repository, some files are generated under `bin/test/`, so I added this folder in `.gitignore`. You can take a look by yourself if you want to:

https://gitpod.io/#https://github.com/sdkman/sdkman-cli

felipecassiors:
@marc0der I think I've got too much excited... 

Please note, I believe the test case `I enter "([^"]*)" and answer "([^"]*)"$'` are not working, and we only caught it now because of the status code check. At least for the `Y` answer. I tried to solve but I was not able to, I didn't find any official reference for the Groovy's `command.execute()` which could help me. So I need your help here. (Just in case you don't know, the command `yes` accept a parameter string, which can be the desired answer such as `yes ${answer} | sdk install java`.

I also believe the post-install hook is not working and we only caught it now.

src/main/bash/sdkman-current.sh
```diff
@@ -25,7 +25,8 @@ function __sdk_current() {
 		if [ -n "$CURRENT" ]; then
 			__sdkman_echo_no_colour "Using ${candidate} version ${CURRENT}"
 		else
-			__sdkman_echo_red "Not using any version of ${candidate}"
+			__sdkman_echo_stop "Not using any version of ${candidate}"
```
felipecassiors, line 28: I added a "Stop!" here.

src/main/bash/sdkman-list.sh
```diff
@@ -28,7 +28,8 @@ function __sdk_list() {
 
 function __sdkman_list_candidates() {
 	if [[ "$SDKMAN_AVAILABLE" == "false" ]]; then
-		__sdkman_echo_red "This command is not available while offline."
+		__sdkman_echo_stop "This command is not available while offline."
```
felipecassiors, line 31: I added a "Stop!" here.

src/main/bash/sdkman-upgrade.sh
```diff
@@ -39,11 +39,14 @@ function __sdk_upgrade() {
 		upgradable="$(__sdkman_determine_upgradable_version "$candidate")"
 		case $? in
 		1)
-			$all || __sdkman_echo_red "Not using any version of ${candidate}"
+			$all || {
+				__sdkman_echo_stop "Not using any version of ${candidate}"
+				return 1
+			}
```
felipecassiors, line 45: I added a "Stop!" here. 

And I'm not sure, but am I preventing the installation of the other candidates if a single fail occur? 

If that's true: I'm doing the same for the next case. So if it's wrong, it should be changed there as well.
   ```
		2)
			echo ""				echo ""
			__sdkman_echo_red "Stop! Could not get remote version of ${candidate}"				__sdkman_echo_stop "Could not get remote version of ${candidate}"
			return 1				return 1
			;;				;;
   ```

If that's not true: I believe we should add a new test case to ensure the installation of one candidate succeeds despite of the fail of another candidate.

src/test/cucumber/mnemonics.feature
```diff
@@ -40,26 +44,30 @@ Feature: Mnemonics
 		When I enter "sdk ug grails" and answer "n"
 		Then I see "Upgrade:"
 		And I see "grails (1.3.9 < 2.4.4)"
+		And the exit code is 0
 
 	Scenario: Shortcut for installing a Candidate Version
 		Given the candidate "grails" version "2.1.0" is not installed
 		And the candidate "grails" version "2.1.0" is available for download
 		And the system is bootstrapped
-		When I enter "sdk i grails 2.1.0" and answer "Y"
```
felipecassiors, line 48: I believe that in this situation, the candidate was not previously installed, so do not have to answer.

src/test/cucumber/offline_mode.feature
```diff
@@ -34,19 +38,23 @@ Feature: Offline Mode
 		And the system is bootstrapped
 		When I enter "sdk offline disable"
 		Then I see "Online mode re-enabled!"
-		When I enter "sdk install grails 2.1.0" and answer "Y"
```
felipecassiors, line 37: I believe that in this situation, the candidate was not previously installed, so do not have to answer.

src/test/cucumber/pre_and_post_installation_hooks.feature
```diff
@@ -7,26 +7,29 @@ Feature: Hooks
...
-		And the candidate "java" version "8u92" is available for download on "Linux"
-		And a "pre" install hook is served for "java" "8u92" on "Linux" that returns successfully
-		And a "post" install hook is served for "java" "8u92" on "Linux" that returns a failure
-		When I enter "sdk install java 8u92"
+		And the candidate "grails" version "2.1.0" is available for download on "Linux"
+		And a "pre" install hook is served for "grails" "2.1.0" on "Linux" that returns successfully
+		And a "post" install hook is served for "grails" "2.1.0" on "Linux" that returns a failure
+		When I enter "sdk install grails 2.1.0"
 		Then I see "Post-installation hook failure"
+		And the exit code is 1
```
felipecassiors, line 35: I believe Java 8u92 and 8.0.101 requires a custom post-installation script, not the default one provided by the Cucumber statement. So I changed to grails 2.1.0, which works. I believe this error was there but we weren't catching it before.

felipecassiors:
@marc0der It seems good to me. I commented everywhere I thought it was important.

.gitignore
```diff
@@ -18,3 +18,4 @@ mongo.json
 out/
 sdkman-cli.iml
 target/
+bin/test/
```
felipecassiors, line 21: See https://github.com/sdkman/sdkman-cli/pull/777#issuecomment-660773714

felipecrs:
Can someone help to review?

przemek-pokrywka (approved):
I was mostly interested in the change that makes SDKMAN exit successfully if something was already installed, which will be super-useful for scripting. From that perspective the changes look alright to me, thanks!

marc0der:
@felipecrs sorry to mess you around like this, but it feels like this PR has drifted off substantially from what it was intended to do, namely _sdk install should be no-op if package already installed_.

I'm all up for some additional improvements but would prefer them in separate smaller PRs. Having small PRs makes it much easier to review and merge. I'd suggest taking a step back and first fixing the actual problem (with it's associated test) so we can get that deployed asap for @przemek-pokrywka to start using. Then once that is done we can do fixing of tests, changing behaviour etc in subsequent fine-grained PRs. wdyt?

przemek-pokrywka:
Hi, thanks both, these changes will certainly be useful for many people, but I'll be good even without them for now, as I needed to remove the sdkman dependency from my project (https://github.com/tsk-tsk/tsk-tsk), because sdkman had too many transitive dependencies (bash, curl, unzip, which, zip), which made it less practical on truly minimal systems (most of the Linux distribution Docker images) than https://get-coursier.io/docs/cli-java.

So it's no longer asap for me, apologies for the shameless plugs, I hope they make it clear why it's no longer asap for me at least.
I'll also applaud the suggestion to make smaller PRs, way more easy to review a small change.

All the best to everybody!

felipecrs:
@marc0der If that's the only way to get this merged, then I have no choice, right?

src/test/cucumber/install_candidate.feature
```diff
@@ -35,8 +35,8 @@ Feature: Install Candidate
 		And the candidate "grails" version "1.3.9" is available for download
 		And the candidate "grails" version "1.3.9" is already installed and default
 		When I enter "sdk install grails 1.3.9"
-		Then I see "Stop! grails 1.3.9 is already installed."
-		And the exit code is 1
+		Then I see "grails 1.3.9 is already installed."
+		And the exit code is 0
```
marc0der, line 39: That's the one :+1:

src/test/cucumber/pre_and_post_installation_hooks.feature
marc0der: The intent of the test was to ensure that pre and post hooks (for Java) are behaving properly. Sorry, I should have been a bit more explicit in the description of this test.

We already have two canned responses (success and failure) by way of `jdk-8.0.101` and `jdk-8.0.111` that should be used instead.

Let's fix this as a separate PR please. It really shouldn't take very long.

src/test/cucumber/service_unavailable.feature
```diff
@@ -63,12 +70,14 @@ Feature: Service Unavailable
...
 		Then I see "Stop! grails 2.1.0 is not available while offline."
+		And the exit code is 1
 
 	Scenario: Install a candidate version that is already installed while Offline
 		Given the candidate "grails" version "2.1.0" is already installed and default
 		And the system is bootstrapped
 		When I enter "sdk install grails 2.1.0"
-		Then I see "Stop! grails 2.1.0 is already installed."
+		Then I see "grails 2.1.0 is already installed."
+		And the exit code is 0
```
marc0der, line 80: :+1:

marc0der (commented):
I do prefer having this as separate PRs as squashing it all together into a single commit will be a horrible mess. It shouldn't take long to extract the two bits into separate PRs and to undo those changes here in this PR. It just leaves a cleaner track of changes in the project's history.

src/main/bash/sdkman-utils.sh
```diff
@@ -103,6 +103,10 @@ function __sdkman_echo_cyan() {
 	__sdkman_echo "36m" "$1"
 }
 
+function __sdkman_echo_stop() {
```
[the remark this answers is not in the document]
felipecrs: It's that old story... [Don't repeat yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

Do you want me to revert?

src/main/bash/sdkman-upgrade.sh
felipecrs: I agree.

src/test/cucumber/pre_and_post_installation_hooks.feature
felipecrs: I agree.

felipecrs:
Thanks @marc0der. I'll split the PR into more.

src/main/bash/sdkman-utils.sh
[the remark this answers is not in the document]
marc0der: Yeah we all know about DRY. In this case I'd say you haven't actually gained anything by shifting a word from the string into the function name. Let's please revert it.

marc0der:
Cool, thanks for understanding @felipecrs. I know I'm a bit extreme about git hygiene, but it does result in a cleaner timeline in the end :smile:

marc0der:
@felipecrs Regarding the broken test for pre/post install hooks. Don't worry about that one, I'll fix it. Shouldn't take me much time and it should save you yours.

marc0der:
@felipecrs Okay, I've ended up using grails like you had it, I was confusing it with the [java installation feature](https://github.com/sdkman/sdkman-cli/blob/master/src/test/cucumber/java_installation.feature) which also exercises the hooks. I'll put it on master so you can rebase.

felipecrs:
@marc0der Alright. I'll try to make it today yet.

felipecrs:
@marc0der ready for review 👍🏻

src/main/bash/sdkman-utils.sh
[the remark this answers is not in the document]
felipecrs: Well, ok, despite it helps to enforce output messages patterns for errors and prevents someone from typing "stop" in a different manner (such as "stop!" and not "Stop!" for example).

marc0der:
Thanks for the contribution!

felipecrs:
@marc0der, any plans about when to release this?

PR #780, opened 2020-07-16
Run sdk env on opening a new terminal

### LINKED ISSUE #779: Bug: .sdkmanrc is not loaded when opening a new terminal
**Bug report**

`.sdkmanrc` is loaded when I cd into a directory. Once there, if I open a new terminal in the same directory it is not loaded.
I'm using another similar tool (https://github.com/direnv/direnv) to export environment variables when _cd-ing_ into a directory and it works well when I open a new terminal. Not really sure how that tool or Sdkman detect this.

**To reproduce**

```
cd /tmp
mkdir foo   
cd foo
sdk env init
.sdkmanrc created.

cd .. && cd -

Using java version 8.0.252-open in this shell.

# Open a new terminal from the same directory (CTRL+SHIFT+T in Linux Mint)
# The new terminal is on the same directory but it doesn't load the env
```

Peek 2020-07-16 10-07

**System info**
- Linux Mint 19.3
- zsh 5.4.2 (x86_64-ubuntu-linux-gnu)
- SDKMAN 5.8.3+506

helpermethod:
Fixes #779

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -145,4 +145,6 @@ if [[ "$sdkman_auto_env" == "true" ]]; then
 
 		[[ -z "$PROMPT_COMMAND" ]] && PROMPT_COMMAND="sdkman_auto_env" || PROMPT_COMMAND="${PROMPT_COMMAND%\;};sdkman_auto_env"
 	fi
+
+  sdkman_auto_env
 fi
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -146,5 +146,5 @@ if [[ "$sdkman_auto_env" == "true" ]]; then
 		[[ -z "$PROMPT_COMMAND" ]] && PROMPT_COMMAND="sdkman_auto_env" || PROMPT_COMMAND="${PROMPT_COMMAND%\;};sdkman_auto_env"
 	fi
 
-  sdkman_auto_env
+	sdkman_auto_env
 fi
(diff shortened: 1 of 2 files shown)


helpermethod:
I will add some tests this in the late evening.

ilopmar:
Yeah, it works!!! :clap:

marc0der:
My biggest concern here is with performance. The `sdk env` command has been notoriously slow, so will this mean that they will experience this lag on every terminal that is opened? Can we benchmark it first?

helpermethod:
@marc0der The `sdk env` command is actually quite fast. The problem are the `curl` invocations in the -cache.sh and -availability.sh files. I will share some benchmarks this evening.

marc0der:
@helpermethod Thanks that would be very helpful to see! If performance isn't an issue then I'm happy to merge this asap.

helpermethod:
> My biggest concern here is with performance. The `sdk env` command has been notoriously slow, so will this mean that they will experience this lag on every terminal that is opened? Can we benchmark it first?

- `sdk env` invocation

image

- `__sdk_env` invocation

image

As you can see, the `__sdk_env` function itself is pretty fast (it would be possible to shave of a few ms but at the expense of code clarity).

helpermethod:
It's interesting where the majority of the time is spent. For this I'm using [gnomon](https://github.com/paypal/gnomon) to get a rough idea about hotspots

```sh
{ PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'; set -x; sdk env; } 2>&1 >/dev/null | gnomon --high 0.1
```

It's easy to see where most of the time is spent

image

helpermethod:
@marc0der Also not that it will only be executed *if* an `.sdkmanrc` exists in the current directory of the newly opened terminal.

marc0der:
> The output makes it pretty obvious where most of the time is spent (with a total running time of 0.9736s).

AFAIK, that should only affect beta channel users as the call is not made every time on the stable channel.

So are you happy for me to merge?

helpermethod:
@marc0der I'll just add a test, ok :)?

helpermethod:
Ready to merge!

marc0der:
Thank you!

PR #782, opened 2020-07-20
Add completion for `sdk` and `sdk current`

### LINKED ISSUE #781: Feature: Bash Completion for SDKMAN

The ergonomics of SDKMAN could be greatly improved by providing autocompletion for commands, subcommands, candidates and versions.

helpermethod:
Fixes #781.


helpermethod:
This is a first, rough iteration. More goodness to come 😄 !

src/main/bash/sdkman-completion.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+_sdk() {
```
marc0der, line 3: How do you feel about calling this top level function something like `_sdk_completion` or something similar? This will keep it distinct from the rest of the sdkman internal functions.

src/main/bash/sdkman-init.sh
```diff
@@ -148,3 +148,5 @@ if [[ "$sdkman_auto_env" == "true" ]]; then
 
 	sdkman_auto_env
 fi
+
+[[ "$bash_shell" == "true" ]] && source sdkman_completion.sh
```
marc0der, line 152: Have you thought about where we will place this completion script? Will it also live in the `bin` folder, or somewhere else like a `contrib` folder. Either way, we would need to ensure that the `sdkman-completion.sh` is extracted into the right location during the _install.sh_ and _selfupdate.sh_ in the sdkman-hooks app.

marc0der:
This is a great start! Can't wait to see where this is going to go!

src/main/bash/sdkman-completion.sh
helpermethod: Good idea!

src/main/bash/sdkman-init.sh
helpermethod: `contrib` would be a new folder? I've just put it into `bin` to make it work locally 😄 . But I can see why `bin` would not be a good place for it.

marc0der: After sleeping on it, I think it would be fine in the `bin` folder, but perhaps explicitly named as `bash-completion.sh`. That would make things a lot simpler. wdyt?

helpermethod: In that case, is it enough to just source `bash-completion.sh`?

helpermethod:
Hi, this is set of completions suited for a first version (aka the low hanging fruit).

Some things I discovered / thought about:

I've tried my hands on `sdk list` but it was dog slow (because of the HTTP call being made). It would probably be a good idea to somewhere store the list of candidates.

The most complex completion will probably be the one for `sdk install`. For that case, we probably need to extend the server API to return candidates based on a prefix to be acceptably fast.

Another problem: there needs to be some way to keep the completions and SDKMAN! itself in sync. I've seen that other commands have integrated that as a subcommand (`rbenv` e.g.). I guess they somewhere declare a shared list / map or something.

Tests are also missing, I would add them as part of the next increment if that would be okay for you 😄 .

helpermethod:
@marc0der Testing the Bash completion turns out to be surprisingly hard 😭 . Any suggestions?

Two problems:

- I have no idea how to insert the double <tab>, I'm also not sure how to capture the output of the completion
- Testing the completion function in isolation is also not possible. While I can set the env variables necessary for comouting the completion, I'm not able to capture the output env variable `COMPREPLY` (because it's only set in the subprocess).

The only option I can think of: having a dedicated function which returns the list of completions as a string, and test that function.

marc0der:
I must be very very honest with you (and this is coming from a hard core TDD guy), but I don't really see the value of having tests for this auto completion. I would simply leave the tests for this 🤷🏻‍♂️

helpermethod:
This is also my conclusion. So, merge?

src/main/bash/sdkman-init.sh
```diff
@@ -148,3 +148,5 @@ if [[ "$sdkman_auto_env" == "true" ]]; then
 
 	sdkman_auto_env
 fi
+
+[[ "$bash_shell" == "true" ]] && source "${SDKMAN_DIR}/src/bash-completion.sh"
```
marc0der, line 152: How about we put this in the `bin` folder instead? Also, can we check that the file is present before we source it, otherwise this will break when we merge to master (we still need to find a reliable way of placing it in the archive and unpacking it onto all our users' systems in the right place).

helpermethod: So, a dedicated file in the bin folder? Will this still be called from the init script?

marc0der: Yes, I'd place it in the bin folder next to the init script, making it easier for it to be sourced. It also means we don't need to do anything with weird relative paths. The archive is bundled together with Gradle, but then extracted using the `install.sh` and `selfupdate.sh` scripts in the [sdkman-hooks](https://github.com/sdkman/sdkman-hooks/tree/master/app/views) service.

marc0der: Here is where the init script is moved in `install.sh`: https://github.com/sdkman/sdkman-hooks/blob/master/app/views/install.scala.txt#L262
Here is the equivalent in selfupdate: https://github.com/sdkman/sdkman-hooks/blob/master/app/views/selfupdate.scala.txt#L101

helpermethod: Hi @marc0der, one last question: The bash-completion.sh script should still be put into the `src` folder, right?

helpermethod:
@marc0der This should be it with regards to the CLI. Not sure why the hooks tests fail. Probably because the changes to the CLI needs to merged first.

marc0der:
No worries, I'll take a look tomorrow when I have some time.

helpermethod:
Found some bugs in the completion file (shortened the code a little and broke some things in the process). Will fix this later in the evening.

mperezi:
Any progress on this?

helpermethod:
@mperezi This is actually already working, at least for the simple cases.

I think the open question was where to put the completion file and how to source it.

@marc0der wdyt, is the issue worth revisiting?

marc0der:
Sorry, this issue must have slipped through the cracks for me.

I don't think it should be placed in the `src` folder as this is purely for bash scripts that need to be sourced on initialisation. How about we place it under `etc`? It isn't strictly a config, but most Linux distros have their completion scripts live here too.

PR #785, opened 2020-07-28
Add `sdk env install` subcommand

### LINKED ISSUE #784: Feature: Add `sdk env install` subcommand

**Feature request**
As discussed some time ago in the Slack `cli-development` channel, it would be nice to have an additional subcommand for `sdk env` which installs the candidates contained in `.sdkmanrc`.

helpermethod:
Fixes #784.

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-all.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
--- a/src/main/bash/sdkman-env.sh
+++ b/src/main/bash/sdkman-env.sh
@@ -25,10 +25,10 @@ function __sdk_env() {
 		__sdkman_env_init "$sdkmanrc"
 		;;
 	install)
-		__sdkman_env_each_line "$sdkmanrc" "__sdk_install"
+		__sdkman_env_each_candidate "$sdkmanrc" "__sdk_install"
 		;;
 	*)
-    	__sdkman_env_each_line "$sdkmanrc" "__sdk_use"
+    	__sdkman_env_each_candidate "$sdkmanrc" "__sdk_use"
     	;;
 	esac
 }
@@ -52,7 +52,7 @@ function __sdkman_env_init() {
 	__sdkman_echo_green "$sdkmanrc created."
 }
 
-function __sdkman_env_each_line() {
+function __sdkman_env_each_candidate() {
 	local -r sdkmanrc="$1"	
 	local -r cb="$2"
 
@@ -64,7 +64,13 @@ function __sdkman_env_each_line() {
 		return 1
 	fi
 
+	local lines=()
+
 	while IFS= read -r line || [[ -n "$line" ]]; do
+		lines+=("$line")
+	done < "$sdkmanrc"
+
+	for line in "${lines[@]}"; do
 		line="${line/\#*/}"
 		line="${line//[[:space:]]/}"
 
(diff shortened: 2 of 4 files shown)


helpermethod:
Hi @marc0der,

in the context of this PR I was thinking about this piece of code

```java
		if ! __sdkman_matches_candidate_format "$line"; then
			__sdkman_echo_red "Invalid candidate format!"
			echo ""
			__sdkman_echo_yellow "Expected 'candidate=version' but found '$line'"

			return 1
		fi
```

I wonder if it actually brings much additional value. The format for `.sdkmanrc` is so simple that probably only very few people actually run into issues. Which makes this pretty much dead code. What do you think?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-6.3-all.zip
```
marc0der, line 3: This causes big downloads on CI, mind keeping this as it was?

helpermethod: Ah, damn, that was definitely not intended! Will remove this ASAP.

helpermethod:
There is now a "higher-order function" __`sdkman_env_each_line()` because `install` and `env` does almost the same. The naming could probably be improved 😄 .

src/main/bash/sdkman-env.sh
```diff
@@ -73,16 +52,36 @@ function __sdkman_generate_sdkmanrc() {
 	__sdkman_echo_green "$sdkmanrc created."
 }
 
-function __sdkman_is_blank_line() {
-	[[ -z "$1" ]]
-}
+function __sdkman_env_each_line() {
```
marc0der, line 55: Should this be called something like `__sdkman_env_each_candidate` instead? Maybe I misunderstood, but won't this function process the entire file for either the `install` or `use` subcommands? The loop actual looping of individual lines happening from line 67 onward.

helpermethod: Yes exactly, thanks!

helpermethod:
There seem to be some problems with the `install` command.

image

helpermethod:
Similar behaviour if the first candidate is made the new default.
image

marc0der:
Here come the complications :smile:

In other words, when `sdk env install` is invoked:
* selfupdate functionality should be omitted
* always answer _no_ to setting new version as default

Is that what you had in mind?

helpermethod:
You were faster than me ;). Damn, was pretty sure this would be a small change 😕 . Any invocation of `read` is problematic because it consumes the next line.

Maybe I could read the whole file in one operation and then just iterate over the (in-memory) list? I guess that should get rid of most of the problems.

image

marc0der:
It feels for me like the approach taken with line-by-line reading is the right one.

What would happen if we set `sdkman_auto_answer` to true temporarily in that function, and make the default `n`? The selfupdate issue seems like a separate concern that might result in some deeper surgery when we decouple it, but there we could live with the know issue as this doesn't happen too often.

helpermethod:
Hi,

lines would still be read line by line

```sh
	while IFS= read -r line || [[ -n "$line" ]]; do
		lines+=("$line")
	done

	for line in "${lines[@]}"; do
		line="${line/\#*/}"
		line="${line//[[:space:]]/}"

		[[ -z "$line" ]] && continue

		if ! __sdkman_matches_candidate_format "$line"; then
			__sdkman_echo_red "Invalid candidate format!"
			echo ""
			__sdkman_echo_yellow "Expected 'candidate=version' but found '$line'"

			return 1
		fi

		"$cb" "${line%=*}" "${line#*=}"
	done < "$sdkmanrc"
```

It's actually moire *in line* (get it 😄 ?) with how I would solve the problem in a high level language like Groovy or Kotlin

```kotlin
File(".sdkmanrc")
    .readLines()
    .map { ... }
    .forEach { ... }
```

The idiomatic way in Bash 4+ would be to use `readarray` / `mapfile` for this, which also reads the whole file into an array.
But this would break compatibility with Bash 3.

helpermethod:
Hm, okay, `sdkman_auto_answer`may still make sense. Not fun being asked if every candidate should become the new default.

helpermethod:
Made the necessary changes. Needs more testing, though.

helpermethod:
How about we introduce another config value like `sdkman_set_default` with default `y`? Then I could overwrite it when calling `__sdk_install`? It would also allow to make `USE` a local variable, with `$sdkman_set_default` as its default value (instead of '').

joehorsnell:
This is great @helpermethod. It would be very useful to get this PR ready and merged as without it, the `env` command is of limited use, IMO.

Need any help getting this PR over the line?

helpermethod:
@joehorsnell I will pick up the issue again as soon as the `sdk env clear` command is merged.

joehorsnell:
Hey @helpermethod, was #825 the only blocker for this PR?

helpermethod:
@joehorsnell There was a regression in one of the last PRs but today I will continue working on the ticket! Stay tuned!

helpermethod:
Implementation already works but needs more tests.

image

helpermethod:
Closing in favor of #839.

helpermethod:
Closing this in favor of #839.

helpermethod:
Closing this in favor of #839.

PR #800, opened 2020-09-19
Add support to Git Bash (switch default)

gayanper:
When working with git bash it ends up with an error saying it cannot remove the directory 'current' adding the -R will solve it for both nix and win.


marc0der:
Thanks for contributing!

gayanper:
Your welcome @marc0der

PR #819, opened 2020-11-10
Update sdkman-help.sh

BFriedmanWorx:
Added clarity to lines 28 and 29.  "default" and "use" do not work with version values, but they do work correctly with identifier values.  So when I run "sdk list java", I see version values and identifier values.  The help page should say "identifier", not "version", for better clarity.  Thank you.


src/main/bash/sdkman-help.sh
```diff
@@ -25,8 +25,8 @@ function __sdk_help() {
 	__sdkman_echo_no_colour "       install   or i    <candidate> [version] [local-path]"
 	__sdkman_echo_no_colour "       uninstall or rm   <candidate> <version>"
 	__sdkman_echo_no_colour "       list      or ls   [candidate]"
-	__sdkman_echo_no_colour "       use       or u    <candidate> <version>"
-	__sdkman_echo_no_colour "       default   or d    <candidate> [version]"
+	__sdkman_echo_no_colour "       use       or u    <candidate> <identifier>"
+	__sdkman_echo_no_colour "       default   or d    <candidate> [identifier]"
```
marc0der, line 29: Version is global to SDKMAN and is true for all candidates (Java is just one of many). We have overloaded the term to accommodate Java in this case, but `Identifier` actually means `version` here.

marc0der:
Apologies for the misunderstanding, but maybe the correct move is to update the sdkman-candidates service and find some other naming for the Java list view columns. This is definitely meant to be the version.

PR #825, opened 2020-11-16
Add `sdk env clear` subcommand

### LINKED ISSUE #824: Feature: Add `sdk env clear` subcommand

**Feature request**
When you are done with a project that uses a custom environment (bootstrapped by `sdk env`) and you walk out of that directory there must be a way of restoring the versions overridden by the environment. This is specially the case when you switch to another project that relies on default versions.

mperezi:
Fixes #824 

Please note that the changes introduced here could use some refactor. However at the time of submitting this work there's an ongoing PR #785 overlapping with this one and addressing just that. Therefore I'd rather wait to see it progress instead of repeating the same work.

### DIFF
--- a/src/main/bash/sdkman-env.sh
+++ b/src/main/bash/sdkman-env.sh
@@ -21,13 +21,15 @@ function __sdk_env() {
 	local -r subcommand="$1"
 
 	case $subcommand in
-		"")    __sdkman_load_env ;;
-		init)  __sdkman_create_env_file ;;
-		clear) __sdkman_clear_env ;;
+		"")    __sdkman_load_env "$sdkmanrc" ;;
+		init)  __sdkman_create_env_file "$sdkmanrc";;
+		clear) __sdkman_clear_env "$sdkmanrc";;
 	esac
 }
 
 function __sdkman_load_env() {
+	local sdkmanrc="$1"
+	
 	if [[ ! -f "$sdkmanrc" ]]; then
 		__sdkman_echo_red "Could not find $sdkmanrc in the current directory."
 		echo ""
@@ -41,6 +43,8 @@ function __sdkman_load_env() {
 }
 
 function __sdkman_create_env_file() {
+	local sdkmanrc="$1"
+	
 	if [[ -f "$sdkmanrc" ]]; then
 		__sdkman_echo_red "$sdkmanrc already exists!"
 
@@ -62,6 +66,8 @@ function __sdkman_create_env_file() {
 }
 
 function __sdkman_clear_env() {
+	local sdkmanrc="$1"
+
 	if [[ -z $SDKMAN_ENV ]]; then
 		__sdkman_echo_red "No environment currently set!"
 		return 1
--- a/src/main/bash/sdkman-env.sh
+++ b/src/main/bash/sdkman-env.sh
@@ -22,20 +22,18 @@ function __sdk_env() {
 
 	case $subcommand in
 	"")
-		__sdkman_env "$sdkmanrc"
+		__sdkman_env
 		;;
 	init)
-		__sdkman_env_init "$sdkmanrc"
+		__sdkman_env_init
 		;;
 	clear)
-		__sdkman_env_clear "$sdkmanrc"
+		__sdkman_env_clear
 		;;
 	esac
 }
 
 function __sdkman_env() {
-	local -r sdkmanrc="$1"
-
 	if [[ ! -f "$sdkmanrc" ]]; then
 		__sdkman_echo_red "Could not find $sdkmanrc in the current directory."
 		echo ""
@@ -63,8 +61,6 @@ function __sdkman_env() {
 }
 
 function __sdkman_env_init() {
-	local -r sdkmanrc="$1"
-
 	if [[ -f "$sdkmanrc" ]]; then
 		__sdkman_echo_red "$sdkmanrc already exists!"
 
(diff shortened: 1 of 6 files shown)


helpermethod:
From my POV #785 can be ignored as there are some currently unsolved problems which prevent implementing this.

When I will revisit #785 I could perform the refactoring within the scope of that ticket (or even create a dedicated ticket for that).

mperezi:
In such a case I will perform the refactor now that I've got myself covered with some tests

src/test/resources/features/per_project_configuration.feature
```diff
@@ -26,3 +26,26 @@ Feature: Per-project configuration
...
+
+	Scenario: The env clear subcommand is issued without an active project configuration
+		Given the system is bootstrapped
+		When I enter "sdk env clear"
+		Then I see "No environment currently set!"
+		And the exit code is 1
+
+	Scenario: The env clear subcommand is issued and the active project configuration is missing
+		Given the system is bootstrapped
+		And a project configuration is active and points to "/tmp"
```
marc0der, line 38: We should be using a random temporary directory here like in other tests to prevent a clash in the residual state.

src/test/groovy/sdkman/specs/EnvCommandSpec.groovy
```diff
@@ -154,6 +154,71 @@ class EnvCommandSpec extends SdkmanEnvSpecification {
...
+
+		when:
+		bash.execute("cd project")
+		bash.execute("cd ..")
+
+		then:
+		bash.output.contains("Restored groovy version to 2.4.6")
+	}
+
+	def "should not execute 'sdk env clear' when entering a subdirectory within the current active configuration"() {
```
marc0der, line 190: Nice one!

src/main/bash/sdkman-init.sh
```diff
@@ -134,15 +134,23 @@ export PATH
...
 
 		chpwd_functions+=(sdkman_auto_env)
 	else
 		function sdkman_auto_env() {
-			[[ "$SDKMAN_OLD_PWD" != "$PWD" ]] && [[ -f ".sdkmanrc" ]] && sdk env
-
-			export SDKMAN_OLD_PWD="$PWD"
+			if [[ "$OLDPWD" != "$PWD" ]] && [[ -f ".sdkmanrc" ]]; then
+				sdk env
+			# environment is set & current dir is not a subdirectory within that environment
```
marc0der, line 150: Mind if we drop this comment?

src/main/bash/sdkman-env.sh
```diff
@@ -34,6 +36,64 @@ function __sdk_env() {
 		return 1
 	fi
 
+	__sdkman_env_each_candidate "$sdkmanrc" "__sdk_use" && 
+		SDKMAN_ENV=$PWD
+}
+
+function __sdkman_create_env_file() {
+	if [[ -f "$sdkmanrc" ]]; then
```
helpermethod, line 44: I would prefer to explicitly pass the `sdkmanrc` variable as a parameter. 

Relying on dynamic scoping makes the code harder to understand because I now also need to take the calling function into account.

@marc0der Wdyt?

marc0der: @helpermethod Totally agree, let's not rely on global scope unless absolutely necessary.

helpermethod: @marc0der Indeed has dynamic scoping similar to Perl, so sth like this works

```bash
x=10

f() {
  local x=20
  g
}

g() {
  echo "$x"
}

f # prints 20
```

But this is such an obscure feature that I would argue against using it (I've basically learned today that this is possible).

src/test/resources/features/per_project_configuration.feature
mperezi: I didn't know about that random temp directory. Much better like that. Done in 108b03f

src/main/bash/sdkman-init.sh
mperezi: No problem. Done in e2b6d63

src/main/bash/sdkman-env.sh
mperezi: TBH I didn't know about dynamic scopes until a couple of days ago 😅  
I thought that maybe in a small module like this one that could be cleaner than passing around the argument to every function.
Anyway it's fixed here 108b03f

marc0der: Very interesting, I wasn't aware of this either. I'd prefer if we just pass values explicitly. Makes the code much easier to follow.

mperezi:
Any stoppers for this? 🙄

helpermethod:
Hi @mperezi!

Looks fine to me! Wdyt @marc0der?

src/main/bash/sdkman-env.sh
```diff
@@ -34,6 +38,68 @@ function __sdk_env() {
...
+
+function __sdkman_env_restore_default_version() {
+	local -r candidate="$1"
+
+	local candidate_dir default_version
+	candidate_dir="${SDKMAN_CANDIDATES_DIR}/${candidate}/current"
+	if __sdkman_is_symlink $candidate_dir; then
+		default_version=$(basename $(readlink ${candidate_dir}))
+		__sdk_use "$candidate" "$default_version" >/dev/null &&
+			__sdkman_echo_no_colour "Restored $candidate version to $default_version (default)"
```
marc0der, line 93: A tiny one, this is informational so could we make this yellow, please? Sorry!

marc0der: Don't worry, I already fixed it ;-)

marc0der:
Great work @mperezi, thanks for the contribution!

Issue #827, opened 2020-11-21
Feature: Add helpful solutions in case of installation/version errors


darshitpp:
- [x] I have read the [CONTRIBUTING guidelines](CONTRIBUTING.md)

Copying from SDKMAN slack channel, please feel free to suggest improvements/edits.

Scenario:
I wanted to try out installing GraalVM JDK on my Linux machine. I have used SDKMan before for installing gradle once and like how easy it was, so decided to use it for this as well.

So I went to the https://sdkman.io/jdks page, where it listed how to install the GraalVM JDK. The command is
```bash
$ sdk install java x.y.z-grl
```
Consequently, I searched the version I wanted to install from the GraalVM website https://www.graalvm.org/downloads/ and tried the command but it returned an error
```bash
$ sdk install java 20.3.0-grl

Stop! java 20.3.0-grl is not available. Possible causes:
 * 20.3.0-grl is an invalid version
 * java binaries are incompatible with Linux64
 * java has not been released yet
```

Unfortunately I still couldn't figure out why it didn't work, so I proceed to manually install it from the GraalVM github page https://github.com/graalvm/graalvm-ce-builds/releases
I then stumbled upon a blog post which described the command to list all the Java versions available with SDKMan (I found this while I was looking for tutorials on Quarkus, btw).
```bash
$ sdk list java
```
And only then I realized that the command listed different versions names than what I was expecting to use while installation.

e.g. I tried `20.3.0` but I was supposed to use `20.3.0.r8`

**Feature request**

* The commands on CLI should point to helpful solutions. The message in this specific case isn't useful, and does not help in solving the problem.
* Update the website to also have the solution to use the sdk list java command before actually trying to install any version

Both the features can have 1 solution, i.e. to describe/suggest to use `sdk list java` in case of such error.

darshitpp:
Closing this issue as the commit has been merged.

PR #828, opened 2020-11-21
#827: Add a helpful tip when incorrect version is provided while installation

darshitpp:
The console will now display additional text describing a helpful command if installation fails due to incorrect verison provided.

### DIFF
--- a/src/main/bash/sdkman-env-helpers.sh
+++ b/src/main/bash/sdkman-env-helpers.sh
@@ -88,6 +88,10 @@ function __sdkman_determine_version() {
 			__sdkman_echo_red " * $version is an invalid version"
 			__sdkman_echo_red " * $candidate binaries are incompatible with $SDKMAN_PLATFORM"
 			__sdkman_echo_red " * $candidate has not been released yet"
+			
+			__sdkman_echo_green "Tip: Try finding a valid version by running the command:"
+			__sdkman_echo_green "sdk list $candidate"
+			__sdkman_echo_green "and rerun with the correct version"
 			return 1
 		fi
 	fi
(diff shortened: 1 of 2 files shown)


src/main/bash/sdkman-env-helpers.sh
```diff
@@ -88,6 +88,10 @@ function __sdkman_determine_version() {
 			__sdkman_echo_red " * $version is an invalid version"
 			__sdkman_echo_red " * $candidate binaries are incompatible with $SDKMAN_PLATFORM"
 			__sdkman_echo_red " * $candidate has not been released yet"
+			
+			__sdkman_echo_green "Tip: Try finding a valid version by running the command:"
```
marc0der, line 92: I would rather suggest that people install relevant auto-completion. There is already one available for ZSH, and another one is in the works for bash. This will be packaged as part of SDKMAN in the near future.

marc0der: Alternatively, a very concise message like:
```
Tip: see all available versions for your platform:
   \$ sdk list $candidate
```
The colour should be yellow as this is an informational message.

darshitpp: > I would rather suggest that people install relevant auto-completion

Auto-completions have the downside of slowing the shell startup, which means a lot of people wouldn't prefer using it. Moreover, a lot of people don't know about auto-completions itself, nor specifically about the SDKMAN auto-completion in ZSH. 

I feel it's a good idea to make SDKMAN usable without any additional dependencies like auto-completion.

> Alternatively, a very concise message like:
> 
> ```
> Tip: see all available versions for your platform:
> $ sdk list $candidate
> ```
>The colour should be yellow as this is an informational message.

Good idea! Thanks for the review! I have done the necessary changes below.

marc0der (approved):
Looks great! Thanks for contributing.

Issue #830, opened 2020-11-29
Feature: SDKMAN to offer the possibility to install Native JDK for Apple Silicon


imrabti:
Now that Apple Silicon computers are publicly available, and there is the first native JDK for this CPU architecture from Azul https://www.azul.com/downloads/zulu-community/?package=jdk, it would awesome that I could install it using SDKMAN.

marc0der:
@imrabti I totally agree, it will be fantastic to have this platform covered. Some work needs to be done on the CLI to accommodate this. We use the output of `uname` and `uname -m` to infer the architecture of a platform on SDKMAN. If you perhaps have such a machine available, please help us by providing the output of those two commands.

imrabti:
@marc0der I have a MBP with M1 I tried the two commands and here is the output :
For `uname` I get `Darwin`
And for `uname -m` I get `arm64`

marc0der:
Does anyone know what a normal Mac reports for `uname -m`? I'm guessing that `uname` also returns `Darwin`. Apologies, I'm a Linux user :smile:

azusa:
Hi @marc0der ,
On normal(Intel) mac, `uname` returns `Darwin` and `uname -m` returns `x86_64` .

pr1metine:
Hi all!
Is there an update on this?

mecca831:
Should be doable by adding a similar check for `DarwinARM64` here: https://github.com/sdkman/sdkman-cli/blob/b8b856f4a589e2447dae8c7ce7fe234a982a5d26/src/main/bash/sdkman-init.sh#L34

However, with Rosetta 2, the x86_64 JDKs will run just fine. From a design point of view, if `DarwinARM64` is added, should it be possible to also list/install x86_64 candidates?

The available candidates would be,

* Darwin -> x86_64 jdks
* DarwinARM64 -> x86_64 AND arm64 jdks

Or maybe just ask the user to set `SDKMAN_PLATFORM="Darwin"` specifically to install x86_64 JDKs?

mecca831:
If it helps, this is the output from macOS Arm64 Zulu JDK by Azul:

```
System.out.println(System.getProperty("os.name") + " " + System.getProperty("os.arch"));
```

```
Mac OS X aarch64
```

lptr:
IIUC this issue now is about choosing the `aarch64` variant of the JDK automatically on Apple silicon. I think the matter is somewhat more complicated than that.

[...]

PR #844, opened 2020-12-21
Add support for macOS arm64

mecca831:
Address #830 by adding "DarwinARM64". No Java candidates will be returned since they are not yet added.


mecca831:
Will be addressed in #846

PR #845, opened 2020-12-21
source only files ending with .sh

smac89:
I use zsh and also make use of a [zmodule](https://github.com/zdharma/zinit#zinit-module) that compiles sourced files for faster execution. This compilation produces binary blobs which end in `.zwc`.

The change proposed here is to make sure sdkman only sources the `.sh` files and nothing else, otherwise it will keep producing errors at startup.

My current work-around is to redirect stderr, but I think adding the `.sh` suffix makes more sense.


marc0der:
Great idea and thanks for contributing, @smac89!

PR #846, opened 2020-12-22
Improve platform inference in shell initialisation.

marc0der:
This is the first of multiple PRs to enable support for the new ARM-based candidate versions. This PR is dependent on some serverside work being done first.

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -30,24 +30,67 @@ if [ -z "$SDKMAN_DIR" ]; then
 fi
 
 # infer platform
-SDKMAN_PLATFORM="$(uname)"
-if [[ "$SDKMAN_PLATFORM" == 'Linux' ]]; then
-	if [[ "$(uname -m)" == 'i686' ]]; then
-		SDKMAN_PLATFORM+='32'
-	elif [[ "$(uname -m)" == 'aarch64' ]]; then
-		SDKMAN_PLATFORM+='ARM64'
-	else
-		SDKMAN_PLATFORM+='64'
-	fi
-fi
+
+function infer_platform() {
+	local kernel
+	local machine
+	
+	kernel="$(uname -s)"
+	machine="$(uname -m)"
+	
+	case $kernel in
+	Linux)
+	  case $machine in
+	  i686)
+		echo "Linux32"
+		;;
+	  X86_64)
+		echo "Linux64"
+		;;
+	  armv7l)
+		echo "LinuxARM32"
+		;;
+	  armv8l)
+		echo "LinuxARM64"
+		;;
+	  aarch64)
+		echo "LinuxARM64"
+		;;
+	  *)
+	  	echo "Linux64"
+	  	;;
+	  esac
+	  ;;
+	Darwin)
+	  case $machine in
+	  X86_64)
+		echo "Darwin64"
+		;;
+	  arm64)    
+		echo "DarwinARM64"
+		;;
+	  *)
+	  	echo "Darwin64"
+	  	;;
+	  esac
+	  ;;
+	*)
+	  echo "$kernel"
+	esac
+}
+
+SDKMAN_PLATFORM="$(infer_platform)"
 export SDKMAN_PLATFORM
+
+__sdkman_echo_debug "Inferred platform: $SDKMAN_PLATFORM"
 
 # OS specific support (must be 'true' or 'false').
 cygwin=false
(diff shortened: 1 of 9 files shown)


src/main/bash/sdkman-init.sh
```diff
@@ -30,24 +30,67 @@ if [ -z "$SDKMAN_DIR" ]; then
...
+	kernel="$(uname -s)"
+	machine="$(uname -m)"
+	
+	case $kernel in
+	Linux)
+	  case $machine in
+	  i686)
+		echo "Linux32"
+		;;
+	  X86_64)
```
pago88, line 47: I'm not sure if this is the case always, but in my system i see the x lowercase: x86-64. Not the end of the world cause the default case catches it, but I still think it should be lowercase to actually match the correct string

pago88 (changes requested):

src/main/bash/sdkman-init.sh
marc0der: Thanks for the feedback @pago88. Would you be able to paste your exact shell output for `uname -s` and `uname -m` please? Also, could I ask you to please drop me a dump of `neofetch` if you have it installed?

pago88: I'm running/coding sdkman inside a docker container, so uname -s is Linux and uname -m is x86_64 (lowercase x). For what it's worth, my host is an intel mac, and uname -m also returns x86_64, lowercase x
Result of neofetch on the container:

       _,met$$$$$gg.          root@759d2e7b5100 
    ,g$$$$$$$$$$$$$$$P.       ----------------- 
  ,g$$P"     """Y$$.".        OS: Debian GNU/Linux 10 (buster) x86_64 
 ,$$P'              `$$$.     Host: BHYVE 1.0 
',$$P       ,ggs.     `$$b:   Kernel: 4.19.121-linuxkit 
`d$$'     ,$P"'   .    $$$    Uptime: 5 hours, 46 mins 
 $$P      d$'     ,    $$P    Packages: 233 (dpkg) 
 $$:      $$.   -    ,d$$'    Shell: bash 5.0.3 
 $$;      Y$b._   _,d$P'      CPU: Intel i5-5257U (2) @ 2.698GHz 
 Y$$.    `.`"Y$$$$P"'         Memory: 1274MiB / 1989MiB 
 `$$b      "-.__
  `Y$$                                                
   `Y$$.
     `$$b.
       `Y$$b.
          `"Y$b._
              `"""

marc0der: Great stuff, thanks for taking the time. I'll make that an explicit check.

src/test/groovy/sdkman/specs/PlatformSpec.groovy
```diff
@@ -0,0 +1,35 @@
...
+		where:
+		kernel 		| machine	| platform
+		"Linux"		| "i686"	| "Linux32"
+		"Linux" 	| "X86_64"	| "Linux64"
+		"Linux"		| "armv7l"	| "LinuxARM32"
+		"Linux"		| "armv8l"	| "LinuxARM64"
+		"Linux"		| "aarch64"	| "LinuxARM64"
+		"Linux"     | ""	    | "Linux64"
+		"Darwin"	| "X86_64"	| "Darwin64"
+		"Darwin"	| "arm64"	| "DarwinARM6"
```
eddumelendez, line 30: missing a `4` at the platform

src/main/bash/sdkman-init.sh
```diff
@@ -30,24 +30,67 @@ if [ -z "$SDKMAN_DIR" ]; then
...
+		echo "LinuxARM64"
+		;;
+	  *)
+	  	echo "Linux64"
+	  	;;
+	  esac
+	  ;;
+	Darwin)
+	  case $machine in
+	  X86_64)
```
mecca831, line 66: Similar to the comment above, the output from my Intel MBP is:

```
$ uname -m
x86_64
```

src/main/bash/sdkman-init.sh
marc0der: That's been fixed. Thanks!

src/test/groovy/sdkman/specs/PlatformSpec.groovy
marc0der: Done!

marc0der:
I've merged this with CLI, but Github seems a bit confused. Closing this PR manually.

PR #851, opened 2021-01-07
Fix rm asking to remove file on some zsh

felipecrs:
I'm trying to fix the following:

```sh-session
❯ sdk install java
==== BROADCAST =================================================================
* 2021-01-07: mvnd 0.3.0 available on SDKMAN! https://git.io/JLxCq
* 2021-01-06: micronaut 2.2.3 available on SDKMAN!
* 2021-01-04: gradle 6.8-rc-5 available on SDKMAN!

Downloading: java 11.0.9.hs-adpt

In progress...

################################################################################################################################ 100.0%################################################################################################################################ 100.0%

Repackaging Java 11.0.9.hs-adpt...

Done repackaging...
rm: remove regular file '/home/vscode/.sdkman/tmp/java-11.0.9.hs-adpt.bin'? y

Installing: java 11.0.9.hs-adpt
Done installing!


Setting java 11.0.9.hs-adpt as default.
```

But I think this might actually be part of the package hook script rather than the sdk cli. Anyway, I believe adding `-f` to the `rm` is always good to prevent such thing.

### DIFF
--- a/bin/download.sh
+++ b/bin/download.sh
@@ -50,6 +50,6 @@ fi
 echo "Extracting archive to: $DESTINATION_DIR"
 unzip -oq "$ARCHIVE" -d "$TMP_DIR"
 mkdir "$DESTINATION_DIR"
-mv "$TMP_DIR"/**/* "$DESTINATION_DIR"
+mv -f "$TMP_DIR"/**/* "$DESTINATION_DIR"
 
 rm -rf "$TMP_DIR"
--- a/src/main/bash/sdkman-flush.sh
+++ b/src/main/bash/sdkman-flush.sh
@@ -25,7 +25,7 @@ function __sdk_flush() {
 		;;
 	version)
 		if [[ -f "${SDKMAN_DIR}/var/version" ]]; then
-			rm "${SDKMAN_DIR}/var/version"
+			rm -f "${SDKMAN_DIR}/var/version"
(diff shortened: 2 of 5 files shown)


marc0der:
Hi @felipecrs, could I ask you to fill out the template that you deleted, please. It makes our lives a lot easier in trying to figure out your issue.

marc0der:
What you are seeing here is the effect of a hook served by the sdkman-hooks service. Please feel free to check that out and apply the fix there.

It would be super helpful to know why it is failing in the first place. Could you please provide your system details, like OS, shell version, and any zsh framework that you are using? My suspicion is that you are using a zsh framework that is messing with the default zsh settings.

felipecrs:
```sh-session
❯ command -v rm
alias rm='rm -i'
```

It's not set in [my dotfiles](github.com/felipecrs/dotfiles), so it must come from some of the extensions, such as OhMyZsh. But it does not matter, the thing is: always use `rm -f` in scripts to make sure it will work for everyone.

felipecrs:
Found it: https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/common-aliases

smola:
This is also a common alias in bash. Amazon Linux 2 comes with it by default for the root user.

marc0der (approved):

marc0der:
@felipecrs thanks for contributing!

PR #859, opened 2021-01-27
Improved error message when candidate version not found

gerrytan:
The message now adds an `sdk install` command to inform user how to correct the error.

This is useful when auto env is enabled, and user not aware their colleague has updated the candidate version. In some situation user isn't even aware of the presence of sdkman. All they see was just `Stop! java 15.0.2.fx is not installed`.

See more discussion about this PR in [this Slack thread](https://sdkman.slack.com/archives/CJK9DMV6V/p1611405973006900)

### DIFF
--- a/src/test/resources/features/default_version.feature
+++ b/src/test/resources/features/default_version.feature
@@ -8,9 +8,12 @@ Feature: Default Version
 		Given the candidate "groovy" version "2.0.5" is a valid candidate version
 		And the system is bootstrapped
 		When I enter "sdk default groovy 2.0.5"
-		Then I see "Stop! Candidate version is not installed. Please run:"
-		And I see ""
-		And I see "$ sdk install groovy 2.0.5"
+		Then I see "Stop! groovy 2.0.5 is not available. Possible causes:"
+		And I see "* 2.0.5 is an invalid version"
+		And I see "* groovy binaries are incompatible with LinuxX64"
+		And I see "* groovy has not been released yet"
+		And I see "Tip: see all available versions for your platform:"
+		And I see "$ sdk list groovy"
 
 	Scenario: Default a candidate version that is installed and not default
 		Given the candidate "groovy" version "2.0.5" is a valid candidate version
--- a/src/test/resources/features/default_version.feature
+++ b/src/test/resources/features/default_version.feature
@@ -8,7 +8,9 @@ Feature: Default Version
 		Given the candidate "groovy" version "2.0.5" is a valid candidate version
 		And the system is bootstrapped
 		When I enter "sdk default groovy 2.0.5"
-		Then I see "Stop! groovy 2.0.5 is not installed."
+		Then I see "Stop! Candidate version is not installed. Please run:"
+		And I see ""
+		And I see "$ sdk install groovy 2.0.5"
 
 	Scenario: Default a candidate version that is installed and not default
 		Given the candidate "groovy" version "2.0.5" is a valid candidate version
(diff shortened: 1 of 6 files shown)


src/test/resources/features/default_version.feature
```diff
@@ -8,7 +8,9 @@ Feature: Default Version
 		Given the candidate "groovy" version "2.0.5" is a valid candidate version
 		And the system is bootstrapped
 		When I enter "sdk default groovy 2.0.5"
-		Then I see "Stop! groovy 2.0.5 is not installed."
+		Then I see "Stop! Candidate version is not installed. Please run:"
+		And I see ""
```
marc0der, line 12: This line can be dropped.

src/test/resources/features/home.feature
```diff
@@ -26,7 +26,9 @@ anything else unless it is an actual error.
 		Given the candidate "grails" version "1.3.9" is available for download
 		And the system is bootstrapped
 		When I enter "sdk home grails 1.3.9"
-		Then I see "Stop! grails 1.3.9 is not installed."
+		Then I see "Stop! Candidate version is not installed. Please run:"
+		And I see ""
```
marc0der, line 30: As before

src/test/resources/features/use_version.feature
```diff
@@ -23,7 +23,9 @@ Feature: Use Version
 		Given the candidate "groovy" version "1.9.9" is not available for download
 		And the system is bootstrapped
 		When I enter "sdk use groovy 1.9.9"
-		Then I see "Stop! groovy 1.9.9 is not installed."
+		Then I see "Stop! Candidate version is not installed. Please run:"
+		And I see ""
```
marc0der, line 27: As before

marc0der (changes requested):
I realised that someone else recently did something [very similar](https://github.com/sdkman/sdkman-cli/blob/f6ee7927c77f351cbe6765e886099b286a841103/src/main/bash/sdkman-env-helpers.sh#L87) for Java installations.

How about we mimic this for the sake of consistent behaviour?

Also, take note that we are using different colouring:
* red for the initial alert message
* yellow for the information that follows.

src/test/resources/features/default_version.feature
gerrytan: TIL, fixed, thanks!

src/test/resources/features/home.feature
gerrytan: Fixed

src/test/resources/features/use_version.feature
gerrytan: Fixed

gerrytan:
> Hi @gerrytan! Firstly, thanks so much for your contribution. This helps everyone so much, so keep them coming!
> 
> I realised that someone else recently did something [very similar](https://github.com/sdkman/sdkman-cli/blob/f6ee7927c77f351cbe6765e886099b286a841103/src/main/bash/sdkman-env-helpers.sh#L87) for Java installations.
> 
> How about we mimic this for the sake of consistent behaviour?
> 
> Also, take note that we are using different colouring:
> 
> * red for the initial alert message
> * yellow for the information that follows.
> 
> Sorry for messing you around like this, I'd forgotten about the tip in the Java installation.
> And again, thanks!

No problem at all, I have pushed this commit to address the above feedback: 8983b488d1f388a6277e2ca9f81eb58f3c7b68f6

marc0der (approved):
Great work, thanks for contributing!

gerrytan:
> Great work, thanks for contributing!

Woohoo! Thanks for your help @marc0der!

Issue #858, opened 2021-01-27
Feature: Revisiting Bash Completion


helpermethod:
- [x] I have read the [CONTRIBUTING guidelines](CONTRIBUTING.md)

**Feature request**
This feature request re-visits the idea to introduce Bash Completion for the most common commands and subcommands.

PR #860, opened 2021-01-27
Add Bash Completion via `sdk completion` subcommand

helpermethod:
This is a first try to implement #858.

It adds the `sdk completion` subcommand which spits out a Bash Completion on stdout.

To use the Bash Completion temporarily a user can just source the output of `sdk completion`

    source <(sdk completion)

To add the Bash Completion permanently, the line `source <(sdk completion)` needs to be added to `.bash_profile` or `.bashrc` (this depends on the operation system).

The command could even output instructions on how to add the Bash Completion permanently as a Bash comment.


marc0der:
This looks great but will only work for bash, is it worth adding a qualifier for the shell?
```
sdk completion [bash|zsh]
```

Also, we probably need to this to both the help command and website.

helpermethod:
1. Added `completion` to the help output.

image

2. Shell identifier is now required, for zsh a message is displayed that completion for zsh is not supported yet

image

src/main/bash/sdkman-completion.sh
```diff
@@ -0,0 +1,78 @@
...
+			__sdkman_completion_zsh
+			;;
+		*)
+			__sdkman_echo_red "No shell provided." && return 1
+	esac
+}
+
+function __sdkman_completion_bash() {
+	\cat <<- 'EOF'
+	_sdk_completion() {
```
marc0der, line 36: How about adding the content of this file to a new folder under `~/.sdkman` called something like `contrib`? That way we can version this file and have better control over it.

marc0der:
I love this, apart from one minor change that could also help with zsh completions going forward.

src/main/bash/sdkman-completion.sh
helpermethod: Good point, here documents are also hard to indent. If anyone gonna make modifications to it, he'll probably mess up the indentation. That anyone would probably be me 😂.

helpermethod: I suppose this needs to be done in the `sdkman-hooks` repository?

helpermethod: Should I also create something like `ext.installContribDir` in `build.gradle`?

helpermethod:
Hi @marc0der , I've put the Bash completion into the `contrib` folder like you suggested, but I'm unsure about how to add the completions to the ZIP.

Should I add the entire folder? Or should I only add the completion files?

marc0der:
I would add the entire folder so that we don't need to do any work when we include zsh completion.

helpermethod:
I've added the `contrib/` folder to the ZIP archive. I suppose updating the `sdkman-hooks` project would be the next step?

marc0der (approved):
Looks fantastic, thank you!

marc0der:
@helpermethod I'll merge this once the work in sdkman-hooks is complete.

helpermethod:
I wonder if I shouldn't just add the existing ZSH completion with some minor changes (the static candidates list e.g. is pretty useless by now).

marc0der:
I think that the [zsh plugin](https://github.com/ohmyzsh/ohmyzsh/tree/master/plugins/sdk) was added to oh-my-zsh by @rgoldberg. Could we reuse that perhaps? wdyt?

helpermethod:
Hi, I'd propose to release this feature with Bash completion first and then handle zsh completion in another ticket.

helpermethod:
You may wonder why `sdk completion` redirects warnings and error messages to STDERR. 

The reason is that the process substitution would pass the output to `source` which would basically throw up

image

With redirecting output to STDERR

image

PR #862, opened 2021-02-01
Add a link to a dedicated SDKMAN! course

adriens:
Hi, at the office I have a new programmer and I figured out that I had to spend some time to make him a dema : how to install and use. Finally I thought that it would be much more interesting to produce a KataCoda course.

Maybe will it interest you. Hopefully you'll like it.

Thanks a lot for your great software, it saves us so much time !!!


marc0der:
Hi @adriens, unfortunately I'm not very keen to link our README to a third party site like KataCoda. Especially since they mandate a signup. If you'd be willing to host your course/tutorial in the SDKMAN GitHub space (in say, the wiki), I'd be more than happy to link to it.

PR #865, opened 2021-02-02
Export SDKMAN_OLD_PWD unconditionally

### LINKED ISSUE #863: Bug: auto env in Bash works only once
**Bug report**
It seems that if a directory contains `.sdkmanrc` then in Bash it works only once (to be precise the version is changed once when entering the dir and then reset when leaving it). Subsequent `cd` doesn't set the version as set in `.sdkmanrc`.

**To reproduce**
In Bash `cd` into directory having `.sdkmanrc`, then `cd -`. The changes of e.g. JDK happen as expected.
Then repeat, nothing happens in Bash.

This is different from zsh, zsh changes the JDK every time when 'visited'.

Expected flow in zsh:

```
krypa% cat /tmp/envTest/.sdkmanrc 
# Enable auto-env through the sdkman_auto_env config
# Add key=value pairs of SDKs to use below
java=15.0.2.hs-adpt
krypa% pwd
/tmp
krypa% cd envTest 

Using java version 15.0.2.hs-adpt in this shell.
krypa% cd -
/tmp
Restored java version to 11.0.9.hs-adpt (default)
krypa% cd -
/tmp/envTest

Using java version 15.0.2.hs-adpt in this shell.
krypa% cd -
/tmp
Restored java version to 11.0.9.hs-adpt (default)
```
Perhaps fault flow with Bash?

```
piotr@krypa:/tmp$ cd envTest/

Using java version 15.0.2.hs-adpt in this shell.
piotr@krypa:/tmp/envTest$ cd -
/tmp
Restored java version to 11.0.9.hs-adpt (default)
piotr@krypa:/tmp$ cd -
/tmp/envTest
piotr@krypa:/tmp/envTest$ cd -
/tmp
piotr@krypa:/tmp$
```

**System info**
- Ubuntu 18.04
- Shell and version (`GNU bash, version 4.4.20(1)-release (x86_64-pc-linux-gnu`))
- The output of `sdk version`: `SDKMAN 5.10.0+617`

helpermethod:
Fixes #863 .

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -190,11 +190,11 @@ if [[ "$sdkman_auto_env" == "true" ]]; then
 		function sdkman_auto_env() {
 			if [[ "$SDKMAN_OLD_PWD" != "$PWD" ]] && [[ -f ".sdkmanrc" ]]; then
 				sdk env
-
-				export SDKMAN_OLD_PWD="$PWD"
 			elif [[ -n $SDKMAN_ENV ]] && [[ ! $PWD =~ ^$SDKMAN_ENV ]]; then
 				sdk env clear
 			fi
+
+			export SDKMAN_OLD_PWD="$PWD"
 		}
 
 		[[ -z "$PROMPT_COMMAND" ]] && PROMPT_COMMAND="sdkman_auto_env" || PROMPT_COMMAND="${PROMPT_COMMAND%\;};sdkman_auto_env"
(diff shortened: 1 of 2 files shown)


helpermethod:
@marc0der I've fixed the bug and added a test which checks that `auto-env` still works after automatically running `sdk env clear` (looks weird with the `cd project; cd ..; cd project` steps but was the only way I could think of).

marc0der (approved):

PR #875, opened 2021-02-17
Add edit command to help output and Bash completion

helpermethod:
This PR adds the `edit` command to the `help` output and the Bash completion.

### DIFF
--- a/contrib/completion/bash/sdk
+++ b/contrib/completion/bash/sdk
@@ -6,7 +6,7 @@ _sdk() {
 
 	case "$previous_word" in
 		sdk) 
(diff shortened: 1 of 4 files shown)


marc0der (approved):

marc0der:
Seeing it like this, it actually doesn't make much sense to me. What are we editing? The intent of this command seems very ambiguous to me. Why don't we call it `config` instead?

helpermethod:
You are absolutely correct! Let's rename it to `config`!

helpermethod:
Hi @marc0der, I've renamed the `edit` command to `config` even it's a little bit outside the scope of the ticket. I hope this is okay for you 🙇 .

marc0der (approved):

PR #878, opened 2021-02-26
Fix: always env clear when leaving dir

reitzig:
The previous logic did not clear the environment when the user switches between two directories with `.sdmanrc`.
Now we always clear (if an environment was set), and _then_ load the new one, if any.

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -178,20 +178,22 @@ export PATH
 if [[ "$sdkman_auto_env" == "true" ]]; then
 	if [[ "$zsh_shell" == "true" ]]; then
 		function sdkman_auto_env() {
+			if [[ -n $SDKMAN_ENV ]] && [[ ! $PWD =~ ^$SDKMAN_ENV ]]; then
+				sdk env clear
+			fi
 			if [[ -f .sdkmanrc ]]; then
 				sdk env
-			elif [[ -n $SDKMAN_ENV ]] && [[ ! $PWD =~ ^$SDKMAN_ENV ]]; then
-				sdk env clear
 			fi
 		}
 
 		chpwd_functions+=(sdkman_auto_env)
 	else
 		function sdkman_auto_env() {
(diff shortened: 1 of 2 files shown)


marc0der (commented):
Is it feasible to test this somehow? Maybe @helpermethod can advise.

reitzig:
I haven't checked out the test rig yet (shame on me), but I imagine it should be possible to write up much like [this one](https://github.com/reitzig/sdkman-for-fish/blob/dev/test/features/autoenv.feature#L57).

marc0der:
That's not actually part of our codebase. You can find the tests that cover the `env` command here:
https://github.com/sdkman/sdkman-cli/blob/master/src/test/groovy/sdkman/specs/EnvCommandSpec.groovy

reitzig:
Sure sure.

Added a test that's red without my proposed changes, and green with them. Hope it maintains the spirit of the test suite.

FWIW, can't say I'm overly impressed by those tests; relying on parts of stdout as a proxy of the desired side-effects feels incomplete (and brittle) to me. :thinking:

helpermethod:
While we're at it:

I wonder if we should also change the logic

    [[ ! $PWD =~ ^$SDKMAN_ENV ]]

to

   [[ $PWD != $SDKMAN_ENV ]]

I think the original idea was to clear only when leaving the (project) root directory, which is nice e.g. in a Maven multi-module project but may not be the expected behaviour in other contexts.

reitzig:
> But this could also be taken care of in another issue.

Please. Seems to me these "other contexts" needs defining; maybe the requirements lead to something like adding a `#: root=true` line to the file format, or whatever. Out of scope here.

helpermethod (approved):
LGTM!

helpermethod:
@marc0der Changes LGTM. Shall we merge?

marc0der (approved):

marc0der:
Thanks for contributing and making those changes!

marc0der:
Would you both mind testing this on the beta channel? If you give the green light, I can deploy it to stable.

helpermethod:
I'll give it a go!

reitzig:
LGTM

image

PR #887, opened 2021-03-26
Fix broken tests on Mac OSX

### LINKED ISSUE #873: Bug: Tests on Mac OSX are not running anymore
**Bug report**
After changes to the platform inference, multiple tests stopped working on Mac OSX.

**To reproduce**

**System info**
- macOS Catalina
-GNU bash, Version 5.0.17(1)-release (x86_64-apple-darwin18.7.0)
- SDKMAN master+647

There seem to be multiple problems which lead to the tests failing. 

1. `UnixUtils.groovy` infers the wrong platform

UnixUtils.groovy expects `System.getProperty("os.name")` to resolve to `Darwin`, while it actually resolves to `Mac OS X`.

```
jshell> System.getProperty("os.name")
$1 ==> "Mac OS X"
```

Interestingly, `uname -s` returns `Darwin` on my machine 😕. 

2. `UnameStub` by default uses platform/kernel `Linux` and machine `X86_64`. This causes problems on Mac OSX, because SDKMAN! then chooses a non-compatible version of `sed`.

This could probably be fixed by replacing the platform-dependent code with shell builtins (this would also slightly improve performance, as less processes need to be forked).

helpermethod:
Fixes #873.

### DIFF
--- a/src/test/groovy/sdkman/specs/PlatformSpec.groovy
+++ b/src/test/groovy/sdkman/specs/PlatformSpec.groovy
@@ -1,9 +1,13 @@
 package sdkman.specs
 
+import sdkman.stubs.UnameStub
 import sdkman.support.SdkmanEnvSpecification
 
 class PlatformSpec extends SdkmanEnvSpecification {
+	UnameStub unameStub
+
 	def setup() {
+		unameStub = UnameStub.prepareIn(new File(sdkmanBaseDirectory, "bin"))
 		sdkmanBashEnvBuilder.withCandidates(["groovy"])
 	}
 
--- a/src/test/groovy/sdkman/specs/PlatformSpec.groovy
+++ b/src/test/groovy/sdkman/specs/PlatformSpec.groovy
@@ -1,13 +1,9 @@
 package sdkman.specs
 
-import sdkman.stubs.UnameStub
 import sdkman.support.SdkmanEnvSpecification
 
 class PlatformSpec extends SdkmanEnvSpecification {
-	UnameStub unameStub
-
 	def setup() {
-		unameStub = UnameStub.prepareIn(new File(sdkmanBaseDirectory, "bin"))
 		sdkmanBashEnvBuilder.withCandidates(["groovy"])
 	}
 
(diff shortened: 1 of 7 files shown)


marc0der:
Awesome work!

I merged a fix to get the build green. Would you mind rebasing on master to get this green?

src/test/groovy/sdkman/support/SdkmanEnvSpecification.groovy
```diff
@@ -11,7 +10,6 @@ abstract class SdkmanEnvSpecification extends BashEnvSpecification {
 	SdkmanBashEnvBuilder sdkmanBashEnvBuilder
 
 	CurlStub curlStub
-	UnameStub unameStub
```
marc0der, line 14: It feels like this should be here because it is a base class for all the specs. wdyt?

src/test/groovy/sdkman/specs/PlatformSpec.groovy
```diff
@@ -1,9 +1,13 @@
 package sdkman.specs
 
+import sdkman.stubs.UnameStub
 import sdkman.support.SdkmanEnvSpecification
 
 class PlatformSpec extends SdkmanEnvSpecification {
+	UnameStub unameStub
```
marc0der, line 7: Is it possible to rely on the base class here if we put it back into the `SdkmanEnvSpecification`? The purpose of that class is, so we don't need to create any fixture like this manually.

helpermethod: The problem is that if you call `UnameStub.prepareIn` in the base class it will always create an empty `uname` file, which will break a lot of tests on non-Linux systems. Calling `prepareIn` only makes sense if you also call `forKernel` and `forMachine` *and* `build`.

helpermethod: Or maybe I misunderstood you 😄 .

helpermethod: The fixture is now created in the base class again.

helpermethod:
Okay, one last try 😄 . Now the `UnameStub` is *always* used but platform / arch / machine are resolved from Java system properties.

marc0der (approved):
Great stuff!

Issue #681, opened 2018-12-13
SDKMAN warns about update while there is none


mhogerheijde:
Please tick one:
- [ ] this is a new Feature Request, a conversation was started in the [Gitter CLI Development Room](https://gitter.im/sdkman/cli-dev).
- [x] this is a new Bug Report or Issue, a conversation was started on [Twitter](https://twitter.com/mhogerheijde/status/1072868313846312964).

Please explain the Issue / Feature Request here:

SDKMAN warned me that there was an update, while there actually wasn't:

```
ATTENTION: A new version of SDKMAN is available...

The current version is 5.7.3+337, but you have 5.7.3+337

Would you like to upgrade now? (Y/n): Y
No update available at this time.
```

I don't know how to reproduce the issue, using the `sdk update` command immediately tells me there are no updates.

![screenshot from 2018-12-12 15-52-08](https://user-images.githubusercontent.com/1466140/49922600-7cb98680-feb1-11e8-83bb-3cd0b0daefeb.png)

See also discussion on twitter: https://twitter.com/mhogerheijde/status/1072868313846312964

Details about my system:

```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.10
Release:        18.10
Codename:       cosmic
$ zsh --version
zsh 5.5.1 (x86_64-ubuntu-linux-gnu)
$ xfce4-terminal --version
xfce4-terminal 0.8.7.4 (Xfce 4.12)
$ tmux -V
tmux 2.7
```

marc0der:
Hi @mhogerheijde 

I see two things that might be the cause of what you're seeing here. Firstly, please check to see that your xfce terminal isn't set to use a login shell. This often causes problems with the state update of the shell. The other thing that might be causing issues is the use of tmux, which could be causing some state bleeding between shells.

And of course, there is the possibility that there might be an interaction between both of these, which might explain the very odd message we are seeing. Shell environment are very fragile, so writing tooling that works across all combinations and permutations is very tricky indeed.

mhogerheijde:
```
$ echo $0
-zsh
```

I think that means that I do use a login-shell? Although i have this option disabled

![screenshot from 2018-12-14 12-07-38](https://user-images.githubusercontent.com/1466140/50000148-f24c5200-ff98-11e8-8b0f-e945b8531e39.png)

[...]

PR #888, opened 2021-03-27
Fix update warn on new shells for update cmd

felipecrs:
Fixes https://github.com/sdkman/sdkman-cli/issues/681.

### DIFF
--- a/src/main/bash/sdkman-main.sh
+++ b/src/main/bash/sdkman-main.sh
@@ -69,11 +69,12 @@ function sdk() {
 	# Various sanity checks and default settings
 	#
 
-	# Check version and candidates cache
+	# Check candidates cache
 	if [[ "$COMMAND" != "update" ]]; then
 		___sdkman_check_candidates_cache "$SDKMAN_CANDIDATES_CACHE" || return 1
-		___sdkman_check_version_cache
 	fi
+	# Check version cache
+	___sdkman_check_version_cache
 
 	# Always presume internet availability
 	SDKMAN_AVAILABLE="true"


felipecrs:
I tested locally and it indeed fixes the issue. How to reproduce the issue:

https://github.com/sdkman/sdkman-cli/issues/681#issuecomment-802210852

marc0der:
Hi @felipecrs, happy to merge this but could you please explain exactly what the effect of your change is? It wasn't very apparent from the linked comment.

Also, if at all possible, it would be great to have a test that validates this fix. I realise that this is probably not feasible.

felipecrs:
Well, in short, `sdk update` command was warning about a new version of `sdkman` available on its execution while there was none.

This issue happened because the `update` command was the only one which was not calling the `___sdman_check_version_cache` (therefore the remote version was not being defined and it caused the sdk update check to misbehave), and I could not find any reason why it was filtered just for this command.

I'll think about how feasible it is to create a test for it.

felipecrs:
Hey @marc0der, it will be a bit complex to create a test case for this because:

1. It would require spawning clean shells
2. Rewrite the `~/.sdkman/vars/` files
3. Depend on an external server to know the latest remote version
4. Make sure the current local version is the same of the dynamic remote version
5. Or, for a negative test, make sure it's not the same as the remote version which would trigger an update

It should be possible somehow, but I don't think it actually worth it. The good news is that no tests were broken.

marc0der (approved):

marc0der:
Thanks for contributing!

PR #889, opened 2021-03-31
Remove platform specific code, make tests more deterministic

helpermethod:
Relates to #873.

After thinking a little bit about it, I had an idea to

* make the tests more deterministic
* remove platform specific code, which prevented the first point

A nice side effect of this change is, that a lot of commands (use, install, current) are also a little faster (around 90ms faster).


marc0der (approved):
I was busy trying the same on my local machine, but you beat me to it! We can also do away with `platform` in the `UnameStub` as it is superseded by the `machine` and `kernel` components.

helpermethod:
> I was busy trying the same on my local machine, but you beat me to it! We can also do away with `platform` in the `UnameStub` as it is superseded by the `machine` and `kernel` components.

I've actually started with removing the platform-specific stuff but then other things broke. Took me a few rounds to get everything working but am now super happy with it :).

PR #897, opened 2021-04-22
Use match instead of BASH_REMATCH when using zsh

### LINKED ISSUE #893: Bug: `use` command no longer working as expected with zsh
**Bug report**
With 5.11.1, setting the default for a candidate (i.e. grails) works.  However, if you change the version using 'sdk use', then the candidate can't be found.  I confirmed this on a fresh install with 5.11.1 and on an upgrade.

**To reproduce**
* curl -s https://get.sdkman.io | bash
* open new terminal
* sdk install grails 4.0.9
* which grails <-- the default will succeed / grails can be found until sdk use is called
* sdk use grails 4.0.9
* which grails <-- this will fail

**System info**
* Mac OS Catalina, latest patches
* zsh - zsh 5.7.1 (x86_64-apple-darwin19.0)
* ohmyzsh shell customization with plugins = (git screen mvn jira github gitfast git-flow docker docker-compose kubectl)
* sdk version => SDKMAN 5.11.1+690

helpermethod:
Fixes #893.

### DIFF
--- a/src/main/bash/sdkman-use.sh
+++ b/src/main/bash/sdkman-use.sh
@@ -39,7 +39,12 @@ function __sdk_use() {
 
 	if [[ $PATH =~ ${SDKMAN_CANDIDATES_DIR}/${candidate}/([^/]+) ]]; then
 		local matched_version
-		[[ "$zsh_shell" == "true"  ]] && matched_version=${match[1]} || matched_version=${BASH_REMATCH[1]}
+
+		if [[ "$zsh_shell" == "true" ]]; then
+			matched_version=${match[1]}
+		else
+			matched_version=${BASH_REMATCH[1]}
+		fi
 
 		export PATH=${PATH//${SDKMAN_CANDIDATES_DIR}\/${candidate}\/${matched_version}/${SDKMAN_CANDIDATES_DIR}\/${candidate}\/${version}}
 	fi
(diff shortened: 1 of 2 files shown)


src/main/bash/sdkman-use.sh
```diff
@@ -38,7 +38,10 @@ function __sdk_use() {
 	__sdkman_set_candidate_home "$candidate" "$version"
 
 	if [[ $PATH =~ ${SDKMAN_CANDIDATES_DIR}/${candidate}/([^/]+) ]]; then
-		export PATH=${PATH//${SDKMAN_CANDIDATES_DIR}\/${candidate}\/${BASH_REMATCH[1]}/${SDKMAN_CANDIDATES_DIR}\/${candidate}\/${version}}
+		local matched_version
+		[[ "$zsh_shell" == "true"  ]] && matched_version=${match[1]} || matched_version=${BASH_REMATCH[1]}
```
marc0der, line 42: Sorry to be a pain, but do you mind if we break this out into a proper `if` construct for consistency with the rest of the codebase?

helpermethod: Yes, this makes sense.

helpermethod: Done!

helpermethod:
@marc0der Replaced the "conditional expressions" with proper if-else statements.

marc0der (approved):
LGTM!

PR #906, opened 2021-04-30
Make the `env install` command set the shell versions

### LINKED ISSUE #905: Bug: `sdk env install` does not also set the current shell versions
**Bug report**
The very useful `sdk env install` command added in #839 does not also set the versions in your shell (ie do an `sdk env`). Looking in the code for the [`env` command](https://github.com/sdkman/sdkman-cli/blob/afccde0db167c7661f8177e9aa57619f13598de4/src/main/bash/sdkman-env.sh#L26) confirms this.

I was initially confused by this - having run `sdk env install`, I assumed that I would have the installed versions set in my shell. Do you consider this a bug and if so, would you accept a fix to make `__sdkman_setup_env` also call `__sdkman_load_env`?

Edit: see discussion below - this behaviour might be better controlled through a command-line agument?

**To reproduce**
```
$ echo 'gradle=7.0' > .sdkmanrc
$ sdk env install
$ gradle --version
```

**System info**
- macOS 10.15.7
- zsh 5.8 (x86_64-apple-darwin19.6.0)
- SDKMAN 5.11.0+644

joehorsnell:
Fixes #905.

### DIFF
--- a/src/main/bash/sdkman-env.sh
+++ b/src/main/bash/sdkman-env.sh
@@ -40,6 +40,7 @@ function __sdkman_setup_env() {
 	fi
 
 	sdkman_auto_answer="true" USE="n" __sdkman_env_each_candidate "$sdkmanrc" "__sdk_install"
+	__sdkman_load_env "$sdkmanrc"
 }
 
 function __sdkman_load_env() {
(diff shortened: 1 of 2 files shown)


helpermethod (approved):
LGTM!

marc0der (approved):
Looks great!

PR #924, opened 2021-06-16
Rename github workflow

marc0der:
Introducing Github actions for the CLI :tada:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,26 @@
+name: Releases
+
+on: 
+  push:
+    tags:
+    - '*'
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Run tests
+      run: ./gradlew clean test --info
+    - name: Generate build number
+      id: buildnumber
+      uses: einaregilsson/build-number@v3 
+      with:
+        token: ${{secrets.GITHUB_TOKEN}}
+    - name: Build artifacts
+      run: ./gradlew -Penv=production -PbuildNumber=${{ steps.buildnumber.outputs.build_number }} clean assemble
+    - uses: ncipollo/release-action@v1
+      with:
+        artifacts: "build/distributions/sdkman-cli-*.zip"
+        token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+    - '*'
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Run tests
+      run: ./gradlew clean test --info
```
helpermethod, line 15: Is clean really needed?

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+    - name: Generate build number
+      id: buildnumber
+      uses: einaregilsson/build-number@v3 
+      with:
+        token: ${{secrets.GITHUB_TOKEN}}
+    - name: Build artifacts
+      run: ./gradlew -Penv=production -PbuildNumber=${{ steps.buildnumber.outputs.build_number }} clean assemble
```
helpermethod, line 22: Why not use the tag name?

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+  push:
+    tags:
+    - '*'
+
+jobs:
+
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
```
helpermethod, line 13: Should probably be `actions/checkout@v2`

helpermethod (commented):
Left a few comments, nothing fancy :).

.github/workflows/release.yml
marc0der: Hey ho, rather safe than sorry :smile:

.github/workflows/release.yml
marc0der: Copy and paste from the action's page, but good one.

.github/workflows/release.yml
marc0der: It's complicated. I didn't want to rework the entire pipeline which depends on the build number. We can't use tags because of the beta channel that substitutes this with `master`, which in turn is not unique. I agree that having this build number fabricated isn't ideal, but babe steps!

helpermethod (approved):

PR #926, opened 2021-06-17
Fix config command when using vscode

felipecrs:
```sh-session
❯ echo $EDITOR
code --wait
❯ sdk config
No default editor configured.
Please set the default editor with the EDITOR environment variable.
```

The fix is to simply remove the double quotes, so the spaces in the env var expand as arguments instead of being interpreted as the path for the executable.

### DIFF
--- a/src/main/bash/sdkman-config.sh
+++ b/src/main/bash/sdkman-config.sh
@@ -17,14 +17,14 @@
 #
 
 function __sdk_config() {
-	local -r editor=${EDITOR:=vi}
+	local -r editor=(${EDITOR:=vi})
 
-	if ! command -v "$editor" > /dev/null; then
+	if ! command -v "${editor[@]}" > /dev/null; then
 		__sdkman_echo_red "No default editor configured."
 		__sdkman_echo_yellow "Please set the default editor with the EDITOR environment variable."
 
 		return 1
 	fi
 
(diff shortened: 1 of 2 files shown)


helpermethod:
Hi @felipecrs,

thanks for the contribution, when I implemented this I didn't anticipate that you could pass arguments to the editor 😄.

Two minor things:

* it would be nice if you could add another test case to (ConfigCommandSpec.groovy)[https://github.com/sdkman/sdkman-cli/blob/master/src/test/groovy/sdkman/specs/ConfigCommandSpec.groovy] (you basically need to add another entry to the `setupEnv ` and `verifyOutput` lists)
* `command -v $editor` fixes the problem but for someone reading the code it's not obvious if leaving out the quotes was intended or an oversight

I would prefer to make it a little bit more explict, something like

    local -r editor=(${EDITOR:=vi}) # expand editor into an array

Then this would work

    if ! command -v "${editor[@]}" > /dev/null

Wdyt?

felipecrs:
That's perfect. Thanks for the suggestion. Done!

helpermethod (approved):
LGTM!

helpermethod:
Great stuff! Thank you for your contribution!

felipecrs:
Awesome!

PR #927, opened 2021-06-17
Add selfupdate_enable config and respect auto_answer when auto updating

### LINKED ISSUE #925: How to disable the check for new versions of sdkman?

**Feature request**

This new option would prevent SDKMAN! from checking for newer versions and asking me to install them when I run it.

felipecrs:
This also deprecates the auto_selfupdate config.

The automatic version check makes the things a lot harder to script in non-attended environments -- such as CI, as it keeps waiting for a response if an updated version is found.

Also, I use [my own tool to manage my updates system widely](https://github.com/felipecrs/dotfiles/blob/12d6701733071553fbe39c03cee3581e676d64c4/dot_local/bin/executable_full-upgrade#L148), which I run automatically. 

Fixes #925
Depends-On: https://github.com/sdkman/sdkman-hooks/pull/28
Depends-On: https://github.com/sdkman/sdkman-website/pull/49

### DIFF
--- a/src/main/bash/sdkman-main.sh
+++ b/src/main/bash/sdkman-main.sh
@@ -153,7 +153,7 @@ function sdk() {
 	fi
 
 	# Attempt upgrade after all is done
-	if [[ "$COMMAND" != "selfupdate" ]]; then
+	if [[ "$COMMAND" != "selfupdate" && "$sdkman_selfupdate_enable" == true ]]; then
 		__sdkman_auto_update "$SDKMAN_REMOTE_VERSION" "$SDKMAN_VERSION"
 	fi
 	return $final_rc
--- a/src/main/bash/sdkman-selfupdate.sh
+++ b/src/main/bash/sdkman-selfupdate.sh
@@ -48,7 +48,7 @@ function __sdkman_auto_update() {
 		__sdkman_echo_no_colour "The current version is $remote_version, but you have $version."
 		echo ""
 
-		if [[ "$sdkman_auto_selfupdate" != "true" ]]; then
+		if [[ "$sdkman_auto_answer" == false ]]; then
 			__sdkman_echo_confirm "Would you like to upgrade now? (Y/n): "
 			read upgrade
 		fi
--- a/src/test/resources/features/self_update.feature
+++ b/src/test/resources/features/self_update.feature
@@ -44,7 +44,7 @@ Feature: Self Update
 
 	Scenario: Automatically Selfupdate
 		Given an outdated initialised environment
-		And the configuration file has been primed with "sdkman_auto_selfupdate=true"
+		And the configuration file has been primed with "sdkman_auto_answer=true"
 		And the system is bootstrapped
 		When I enter "sdk help"
 		Then I see "A new version of SDKMAN is available..."
@@ -54,13 +54,21 @@ Feature: Self Update
 
 	Scenario: Do not automatically Selfupdate
 		Given an outdated initialised environment
-		And the configuration file has been primed with "sdkman_auto_selfupdate=false"
+		And the configuration file has been primed with "sdkman_auto_answer=false"
 		And the system is bootstrapped
 		When I enter "sdk help" and answer "n"
 		Then I see "A new version of SDKMAN is available..."
 		And I see "Would you like to upgrade now? (Y/n)"
 		And I see "Not upgrading today..."
 		And I do not see "Successfully upgraded SDKMAN."
+
+	Scenario: Do not check and prompt for update
+		Given an outdated initialised environment
+		And the configuration file has been primed with "sdkman_selfupdate_enable=false"
+		And the system is bootstrapped
+		When I enter "sdk help"
+		Then I do not see "A new version of SDKMAN is available..."
+		And I do not see "Would you like to upgrade now? (Y/n)"
 
 	Scenario: Bother the user with Upgrade message once a day
 		Given an outdated initialised environment


marc0der:
Hi, @felipecrs. To clarify, do you mean to prevent the self-update check from happening in a CI environment, or do I misunderstand your intent? If this is indeed the problem you are facing, we already have a config variable called `sdkman_auto_selfupdate` to cover this.

felipecrs:
Hi, @marc0der. The `sdkman_auto_selfupdate=false` configuration makes the update not to install automatically. But still, `sdk` checks for updates in almost every command, and prompts for the user to update if a newer version is found.

The new configuration `sdkman_version_check=false` will prevent `sdk` from checking for updates automatically and thus, also prevent the update prompt from appearing as well.

Different things, right?

felipecrs:
This issue can be replicated:

```console
❯ docker run -it --rm mcr.microsoft.com/vscode/devcontainers/java:0.201.5-15 bash -c 'echo "sdkman_auto_selfupdate=false" > $SDKMAN_DIR/etc/config && touch -d 20120101 $SDKMAN_DIR/var/delay_upgrade && exec bash -ic "sdk version"'
We periodically need to update the local cache. Please run:

  $ sdk update

==== BROADCAST =================================================================
* 2021-06-18: mvnd 0.5.2 available on SDKMAN! https://git.io/Jn0Kl
* 2021-06-17: micronaut 3.0.0-M2 available on SDKMAN!
* 2021-06-15: micronaut 2.5.6 available on SDKMAN!

SDKMAN 5.11.2+698


ATTENTION: A new version of SDKMAN is available...

The current version is 5.11.5+713, but you have 5.11.2+698.

Would you like to upgrade now? (Y/n): ^C

```

marc0der:
Hey @felipecrs, I understand the problem now. Could we make the naming of this config a little clearer to reflect the _what_, not the _how_? How about calling the config `sdkman_selfupdate_disable` to make it very clear what the behaviour is? This will also require changes to be made in the sdkman-hooks service. I would also like to have a test to cover this if at all possible.

felipecrs:
Sure! But isn't the existence of both `sdkman_auto_selfupdate` and `sdkman_selfupdate_disable` a bit confusing? Just a thought: what if we extend `sdkman_auto_selfupdate` to include what I'm introducing?

marc0der:
The name `sdkman_auto_selfupdate` implies that you are making the self-update automatic (or not). This is distinctly different from not doing a self-update at all. I prefer this as a separate configuration, and we should document both clearly on the website.

Incidentally, the associated changes in the sdkman-hooks service need to be made in the [install](https://github.com/sdkman/sdkman-hooks/blob/master/app/views/install.scala.txt) and [selfupdate](https://github.com/sdkman/sdkman-hooks/blob/master/app/views/selfupdate.scala.txt) scripts.

felipecrs:
> This is distinctly different from not doing a self-update at all

But the new configuration is not removing this ability, it's available there with `sdk selfupdate` command. Actually, it's disabling the "auto selfupdate", and that's why I think it would be less confusing to just extend its capabilities instead of creating a new config.

marc0der:
This really is quite different, and most certainly a feature used by many people. Think about developers who are using this on their machines (as opposed to in a CI environment) who don't want to be asked if the next update should/shouldn't be installed. For this very scenario, we have auto-update.

felipecrs:
What about this existing config:

```sh-session
# make sdkman non-interactive, preferred for CI environments
sdkman_auto_answer=true|false
```

Should not it be respected? If it should, there would not be a point for `sdkman_auto_selfupdate` to exist.

felipecrs:
So I believe the optimal solution would be to:

1. Extend `sdkman_auto_answer=true` to include the current capabilities of `sdkman_auto_selfupdate=true`
2. Change the `sdkman_auto_selfupdate` config to do what the new config being introduced in this PR would do.

marc0der:
I totally agree with what you are proposing, but the name of `sdkman_auto_selfupdate` is completely wrong and will confuse many people. What you are describing should be called `sdkman_selfupdate_enable`. So we should deprecate the old `sdkman_auto_selfupdate`, favouring the new config and behaviour you proposed.

felipecrs:
@marc0der good! I will work on it.

felipecrs:
> BTW, the "auto" in sdkman_auto_selfupdate refers to the fact that it is an unattended update.

Well, it could be what it was intended to mean, but still `auto` tends to mean more `automatic` than `unattended`.

In context, you can turn on/off the `sdkman` `automatic` `self-update` (`sdkman_auto_selfupdate`) with `true/false`.

On the other hand: `sdkman_selfupdate_enable` has nothing to reference the "automatic" part of the process (and it needs to be referenced, as the normal selfupdate (`sdk selfupdate` command) will remain functional even though `sdkman_selfupdate_enable=false`.

Anyway, I'm just discussing to make sure we introduce a good config name to prevent future refactors/breaking changes.

marc0der:
I'm happy with `sdkman_selfupdate_enable` and will merge your PR once this is in place. However, remember that we need those two sdkman-hooks scripts updated before we can merge this one, and we will need appropriate changes made on the usage page of the website.

marc0der:
@felipecrs any update on this one?

felipecrs:
@marc0der I believe all the changes are ready to review.

src/main/bash/sdkman-main.sh
```diff
@@ -153,8 +153,8 @@ function sdk() {
 	fi
 
 	# Attempt upgrade after all is done
-	if [[ "$COMMAND" != "selfupdate" ]]; then
-		__sdkman_auto_update "$SDKMAN_REMOTE_VERSION" "$SDKMAN_VERSION"
+	if [[ "$COMMAND" != "selfupdate" && "$sdkman_selfupdate_enable" != "false" ]]; then
+		__sdkman_auto_update "$SDKMAN_REMOTE_VERSION" "$SDKMAN_VERSION"2
```
marc0der, line 157: Is this `2` a typo?

marc0der (commented):
Looking great, just one comment on this PR.

Also, we require the changes to be made in the sdkman-hooks service.

src/main/bash/sdkman-init.sh
```diff
@@ -140,6 +140,9 @@ if [[ ! -f "${SDKMAN_DIR}/var/delay_upgrade" ]]; then
 	touch "${SDKMAN_DIR}/var/delay_upgrade"
 fi
 
+# set version-check
+if [[ -z "$sdkman_version_check" ]]; then sdkman_version_check=true; fi
```
marc0der, line 144: We don't need this defensive code as it should be set in the config file.

src/main/bash/sdkman-main.sh
```diff
@@ -153,8 +153,8 @@ function sdk() {
 	fi
 
 	# Attempt upgrade after all is done
-	if [[ "$COMMAND" != "selfupdate" ]]; then
-		__sdkman_auto_update "$SDKMAN_REMOTE_VERSION" "$SDKMAN_VERSION"
+	if [[ "$COMMAND" != "selfupdate" && "$sdkman_selfupdate_enable" != "false" ]]; then
```
marc0der, line 156: This double negative seems strange. Why not `sdkman_selfupdate_enable == "true"`?

src/main/bash/sdkman-selfupdate.sh
```diff
@@ -48,7 +48,7 @@ function __sdkman_auto_update() {
 		__sdkman_echo_no_colour "The current version is $remote_version, but you have $version."
 		echo ""
 
-		if [[ "$sdkman_auto_selfupdate" != "true" ]]; then
+		if [[ "$sdkman_auto_answer" != "true" ]]; then
```
marc0der, line 51: Double negative again. How about `$sdkman_auto_answer == false`?

src/main/bash/sdkman-init.sh
felipecrs: Absolutely. My bad.

src/main/bash/sdkman-main.sh
felipecrs: I'm 100% against assuming everyone will have this set in their config files. For example, I manage my dotfiles using another program, so I'm the one who makes changes to it. If later in the future sdkman introduces new config variables, I would miss it. 

I think 2 things are different:

1. Default settings 
2. User preferences (the ~/.sdkman/etc/config file)

And I don't think sdkman should make changes unattended(ly) to the user preferences, but only to the default settings. Just like Git (~/.gitconfig) or any other tool.

And one way to accomplish this behavior is doing such negative comparison:

It will only work if the user setted to false, explicitly, otherwise it will have the truthy behavior (default=true).

src/main/bash/sdkman-selfupdate.sh
felipecrs: Same as above. To prevent issues when the user do not have such entry in its config file.

src/main/bash/sdkman-main.sh
felipecrs: Wow, it is. My bad.

felipecrs:
@marc0der I made the changes. Most of them was due to my lack of attention, while the others I replied you in their own conversation.

src/main/bash/sdkman-main.sh
marc0der: This is a huge leap from how SDKMAN currently works, and I don't think we should shoehorn it into this PR. If we want to go down this route, it should be done consistently for all the configurations, and it should be done as a separate PR.

For this PR, I expect it to be done as it always was. I suggest you open a new feature request for the above.

felipecrs: Cool. That works for me.

felipecrs:
@marc0der all done.

marc0der (approved):

PR #928, opened 2021-06-18
Fix selfupdate when beta channel is not set

felipecrs:
I don't know why this issue started to happen, maybe something changed in the SDKMAN API, but every time I tried to run `sdk selfupdate`, it was failing with some HTML output.

Debugging it, I noticed that it was trying to `curl` `api.sdkman.io/selfupdate?beta=`, and the failure was about beta not being set.

This makes the beta_channel configuration set to `false` by default.


helpermethod:
If you type `sdk config`, do you see an entry for `sdkman_beta_channel`?

Could it be that you've accidentally that entry?

felipecrs:
I checked that. Actually, if I write beta_channel=false in my config the issue is gone as well.

I believe the issue can be replicated anywhere, I just don't actually know how to install an older version of sdkman to simulate an upgrade.

marc0der:
Hi @felipecrs, nothing has really changed in the backend that I'm aware of to cause this behaviour. Both the install and self-update scripts ensure that these config variables are set. This code has never really changed. I guess that you might have deleted it from your config by accident. Might that be possible?

Be that as it may, let's keep an eye on it to see if anyone else experiences this problem. If there really is an issue, I would prefer to fix the root cause rather than adding such defensive code.

felipecrs:
> I guess that you might have deleted it from your config by accident. Might that be possible?

@marc0der, is it mandatory to be set in my config? I thought that I could omit the configs in which I would like to use the default.

felipecrs:
This error can be replicated:

````console
❯ docker run -it --rm mcr.microsoft.com/vscode/devcontainers/java:0.201.5-15 bash -c 'rm -f $SDKMAN_DIR/etc/config && exec bash -ic "sdk selfupdate"'
We periodically need to update the local cache. Please run:

  $ sdk update

bash: line 1: syntax error near unexpected token `newline'
bash: line 1: `<!DOCTYPE html>'
```

marc0der:
@felipecrs I really don't think we need to document this. Nobody is deleting the configs from their config files, and you would be the first person ever to report a problem about this. My best advice: just leave them as they were intended to be, then nothing will break. We even go so far as to ensure that all configs are present in the _/etc/config_ on every self-update.

felipecrs:
The other option is to check if the variable is defined in the selfupdate function, without having to set it in the init. Should I change the implementation of the fix?

felipecrs:
@marc0der can you please review again? I'm using a less intrusive approach now.

Issue #923, opened 2021-06-15
syntax error near unexpected token '<'


ssiwek-dxc:
Executing command "source /root/.sdkman/bin/sdkman-init.sh" fails with:

source /root/.sdkman/contrib/completion/bash/sdk
/root/.sdkman/contrib/completion/bash/sdk: line 37: syntax error near unexpected token `<'
/root/.sdkman/contrib/completion/bash/sdk: line 37: `			done < <(curl --silent "${SDKMAN_CANDIDATES_API}/candidates/all")'

I believe that it's caused by using old bash (from centos-7) which don't support process substitution:

$ bash --version
GNU bash, version 4.2.46(2)-release (x86_64-redhat-linux-gnu)

I encountered this bug in sdkman-5.11.5+713. I believe that previous versions were OK.

ssiwek-dxc:
I think that such commit could fix this issue https://github.com/ssiwek-dxc/sdkman-cli/commit/136d19a9bf5790ee32a6871a7019537b895f3ab0

Not tested.

marc0der:
You can stop this from occurring by disabling the auto-completion.
Set `sdkman_auto_complete=false` in the _.sdkman/etc/config_ file.

ssiwek-dxc:
Yeah, I thought about such thing - but truncating completion script was my first idea

jjkoehorst:
Luckily it is only a single line that needs to be added to the docker build file

    RUN echo "sdkman_auto_complete=false" >> "$HOME/.sdkman/etc/config"

marc0der:
It's an acceptable workaround, but maybe @helpermethod could come up with a solution for this that also works with older versions of `bash` and `sh`.

helpermethod:
@marc0der Will have a look at it!

pavan-onkar-exa:
Wow even I noticed it.
@jjkoehorst Your solution worked smoothly for me !!

helpermethod:
Hi @ssiwek-dxc, your proposal looks good to me. Would you like to submit this as a PR?

kenyee:
Bit me too :-(
Any idea when the PR will be merged?  Looks like it's failing CI though...
Wondering if I should add the workaround...

kenyee:
Also, is there a way to version lock SDK man installs?
This broke on our CI...never expected installing SDK man would fail because we have Centos7 images...

marc0der:
[...]

PR #935, opened 2021-06-21
fix for bash-4 (#923)

ssiwek-dxc:
bash-4 does not support process substitution


marc0der (approved):
LGTM

marc0der:
Thanks for contributing!

PR #942, opened 2021-07-19
fix for bash-4 (sdkman#923)

ssiwek-dxc:
bash-4 does not support process substitution.
Another occurence.

### DIFF
--- a/contrib/completion/bash/sdk
+++ b/contrib/completion/bash/sdk
@@ -72,9 +72,10 @@ __sdkman_complete_candidate_version() {
 			done
 			;;
 		install)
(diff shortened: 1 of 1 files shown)


marc0der (approved):
Awesome, thank you!

ssiwek-dxc:
argh - you are so quick

marc0der:
That's deployed on beta. Let me know if it works for you, and I'll release a patch to stable @ssiwek-dxc.

ssiwek-dxc:
It was not too easy to test this beta release - but I believe that now it's OK

PR #957, opened 2021-08-22
Reshape binary distribution to be SDK compliant

marc0der:
This PR reshapes the binary zip so that it may be extracted in-place during the installation process.

- Extract custom archive tasks to gradle folder.
- Prepare and consume files in new stage build folder.
- Add test task dependency on script prep tasks.
- Silence noisy logs.


helpermethod (approved):
LGTM!

PR #960, opened 2021-08-31
Save HTTP response headers under ${SDKMAN}/var/metadata

hgeraldino:
Store the HTTP response headers for each downloaded candidate. 

These headers will be used in future PRs to extract metadata such as: checksums, file type, file name, etc.

### DIFF
--- a/src/main/bash/sdkman-install.sh
+++ b/src/main/bash/sdkman-install.sh
@@ -120,11 +120,15 @@ function __sdkman_download() {
 	version="$2"
 
 	archives_folder="${SDKMAN_DIR}/archives"
+	metadata_folder="${SDKMAN_DIR}/var/metadata"
+	mkdir -p ${metadata_folder}
+	
 	if [ ! -f "${archives_folder}/${candidate}-${version}.zip" ]; then
 		local platform_parameter="$(echo $SDKMAN_PLATFORM | tr '[:upper:]' '[:lower:]')"
 		local download_url="${SDKMAN_CANDIDATES_API}/broker/download/${candidate}/${version}/${platform_parameter}"
 		local base_name="${candidate}-${version}"
 		local zip_archive_target="${SDKMAN_DIR}/archives/${candidate}-${version}.zip"
+		local headers="${metadata_folder}/${base_name}.headers"
 
 		# pre-installation hook: implements function __sdkman_pre_installation_hook
 		local pre_installation_hook="${SDKMAN_DIR}/tmp/hook_pre_${candidate}_${version}.sh"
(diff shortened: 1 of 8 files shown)


src/test/resources/features/flush.feature
```diff
@@ -42,3 +44,9 @@ Feature: Flush
 		When I enter "sdk flush temp"
 		Then no "res-1.2.0.zip" file is present in temporary storage
 		And I see "1 archive(s) flushed"
+
+	Scenario: Clear out the metadata
+		Given a headers file "grails-1.3.9.headers" in metadata directory
+		When I enter "sdk flush metadata"
+		Then no metadata is cached
+		And I see "1 archive(s) flushed"
```
marc0der, line 52: Should this read something like "Metadata was flushed"?

src/test/groovy/sdkman/steps/installation_steps.groovy
```diff
@@ -96,3 +96,15 @@ And(~'^I have configured "([^"]*)" to "([^"]*)"$') { String configName, String f
 And(~/^the exit code is (\d+)$/) { Integer rc ->
 	assert bash.getStatus() == rc
 }
+
+And(~'^the response headers file is created for candidate "([^"]*)" and version "([^"]*)"$') { String candidate, String version ->
+	def headersFile = "${metadataDir}/${candidate}-${version}.headers" as File
+	if (!headersFile.exists()) println bash.output
```
marc0der, line 102: The printing of the bash output probably shouldn't happen here. I think it is already done elsewhere in the bash environment helper.

src/test/groovy/sdkman/steps/installation_steps.groovy
```diff
@@ -96,3 +96,15 @@ And(~'^I have configured "([^"]*)" to "([^"]*)"$') { String configName, String f
...
+	assert headersFile.exists()
+}
+
+And(~'^no response headers are written for candidate "([^"]*)" and version "([^"]*)"$') { String candidate, String version ->
+	def headersFile = "${metadataDir}/${candidate}-${version}.headers" as File
+	if (headersFile.exists()) println bash.output
```
marc0der, line 108: Same as above.

marc0der (commented):
This is awesome work! I've left some very minor comments to be addressed.

src/test/resources/features/flush.feature
hgeraldino: This option is calling the same function (`__sdkman_cleanup_folder`) as the other option of the flush command (archives, tmp) so the message logged is the same

Not sure if I want to implement a new function

src/test/groovy/sdkman/steps/installation_steps.groovy
hgeraldino: removed

src/test/groovy/sdkman/steps/installation_steps.groovy
hgeraldino: removed

src/main/bash/sdkman-install.sh
```diff
@@ -120,11 +120,15 @@ function __sdkman_download() {
 	version="$2"
 
 	archives_folder="${SDKMAN_DIR}/archives"
+	metadata_folder="${SDKMAN_DIR}/var/metadata"
+	mkdir -p ${metadata_folder}
```
marc0der, line 124: I think this is fine for now, but the creation of this folder can probably be performed by the install and selfupdate hooks in the future.

marc0der (approved):
LGTM, thanks @hgeraldino! Are you happy with these initial changes @helpermethod?

helpermethod (approved):
LGTM! Great work!

Issue #938, opened 2021-06-28
Feature: Skip security check of the Zsh completion system like OMZ does


borrible13th:
- [x ] I have read the [CONTRIBUTING guidelines](CONTRIBUTING.md)

**Feature request**
In a multiuser environment, e.g. macOS Big Sur, Zsh's security check by `compinit` warns about insecure directories and files. As a current user, you are often not either root or the owner of these directories and files. No chance to silence the warning and to use all insecure files and directories without asking.

Please introduce a variable to disable the security check by calling `compinit -u -C` instead. For example Oh-My-Zsh uses `ZSH_DISABLE_COMPFIX="true"`.

marc0der:
Hi @borrible13th, feel free to raise a PR to disable this check. I'd be happy to merge it.

rcousens:
In reference to this: https://zsh.sourceforge.io/Doc/Release/Completion-System.html

Are you okay with compinit being launched with -C by default (to skip the security checks) or do you want it controlled by a variable?

I went in and modified sdkman-init.sh to "compinit -C" for zsh to get rid of the annoying prompts.

danieldietrich:
+1 `compinit -C`

marc0der:
As before, the onus is on you. I'll happily accept a PR for this @danieldietrich @rcousens.

danieldietrich:
@marc0der 

> I'll happily accept a PR for this @danieldietrich

Thanks! PR ready for review

marc0der:
This has been deployed to stable so closing the issue. Thanks again for contributing!

PR #964, opened 2021-09-04
Considering env var ZSH_DISABLE_COMPFIX during initialization

danieldietrich:
This PR provides a fix for #938 by considering the environment variable `ZSH_DISABLE_COMPFIX`.

By default, on initialization `compinit` is called like before. That means `compaudit` is called and security checks are performed. This might block the shell, waiting for user input. Especially this is not desired in the presence of non-interactive shells. Another scenario, in which security checks might not be desired is a multi-user system. For security reasons, `compinit` requires completion directories to be owned by the same user who runs the command.

If we set `ZSH_DISABLE_COMPFIX="true"`, then `compinit -u -C` is called. Effectively, `-C` will skip all security checks and `-u` will use all files found for completion (see [compinit](https://github.com/zsh-users/zsh/blob/82ff9f24f170eea7daa935fdaa09ab75a2f277ff/Completion/compinit#L67-L72)). This is considered best practice in [oh-my-zsh](https://github.com/ohmyzsh/ohmyzsh/blob/28ee5846bbac56926241fc5ef9f116161ceb6f23/oh-my-zsh.sh#L80-L81).

**Note:** this might need to be mentioned in the [sdkman.io docs](https://github.com/sdkman/sdkman-website/blob/master/app/views/install.scala.html)

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -181,7 +181,11 @@ if [[ "$sdkman_auto_complete" == 'true' ]]; then
 		# initialize zsh completions (if not already done)
 		if ! (( $+functions[compdef] )) ; then
 			autoload -Uz compinit
-			compinit
+			if [[ $ZSH_DISABLE_COMPFIX == 'true' ]]; then
+				compinit -u -C
+			else
+				compinit
+			fi
 		fi
 		source "${SDKMAN_DIR}/contrib/completion/zsh/sdk"
 		__sdkman_echo_debug "ZSH completion script loaded..."


src/main/bash/sdkman-init.sh
```diff
@@ -181,7 +181,13 @@ if [[ "$sdkman_auto_complete" == 'true' ]]; then
 		# initialize zsh completions (if not already done)
 		if ! (( $+functions[compdef] )) ; then
 			autoload -Uz compinit
-			compinit
+			if [[ $ZSH_DISABLE_COMPFIX == 'true' ]]; then
+			    # the user decided to skip compaudit security checks
```
marc0der, line 185: I don't think it's necessary for this comments.

src/main/bash/sdkman-init.sh
```diff
@@ -181,7 +181,13 @@ if [[ "$sdkman_auto_complete" == 'true' ]]; then
...
+				compinit -u -C
+			else
+				# perfom compaudit security checks (may wait for user input)
```
marc0der, line 188: Ditto

danieldietrich:
@marc0der thanks for the hints! I applied the changes and force-pushed the new version. Please let me know if there is sth else.

marc0der (approved):
LGTM!

marc0der:
Thanks for contributing this!

marc0der:
This is now deployed in the beta channel. Please let me know if it behaves as expected and I'll be happy to release to stable.

danieldietrich:
@marc0der I can confirm that it works on the beta channel. Here are my tests:

1. ✅  On the lates stable version, I received a compinit warning (this also happened in the presence of the env var `ZSH_DISABLE_COMPFIX="true"`, I have no screenshot for that).

Screen Shot 2021-09-05 at 17 05 19

2. ✅  On the beta channel, we observe the same behavior when the env var `ZSH_DISABLE_COMPFIX` is not set to `"true"`. This is expected because we did not want to change the 'default' behavior.

Screen Shot 2021-09-05 at 17 14 37

3. ✅  The beta version now skips the compinit security checks if we set the env var `ZSH_DISABLE_COMPFIX="true"`. Especially this removes the initial input dialog when opening a new zsh shell, which blocked the terminal.

Screen Shot 2021-09-05 at 17 09 22

marc0der:
Great, thanks for confirming. I'll release this to stable shortly.

danieldietrich:
Thank you, your OS project really works like clockwork! 👍

PR #968, opened 2021-09-10
Display a warning when Bash is run in POSIX mode

helpermethod:
When SDKMAN! is run in [POSIX mode](https://www.gnu.org/software/bash/manual/html_node/Bash-POSIX-Mode.html), a warning should be displayed and Bash completion turned off.


helpermethod:
Closed in favor of #978.

PR #978, opened 2021-09-18
Make Bash completion work with POSIX mode

helpermethod:
Tests are now run with the `--posix` flag by default, to detect cases where something would break in POSIX mode.

Strangely enough, POSIX mode does not disable all of Bash's functionality, so things like `here strings` work. And on Bash 5.x, even process substitution works 😕.

### DIFF
--- a/contrib/completion/bash/sdk
+++ b/contrib/completion/bash/sdk
@@ -73,7 +73,7 @@ __sdkman_complete_candidate_version() {
 			done
 			;;
 		install)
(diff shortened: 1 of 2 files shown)


marc0der (commented):
Looks good, although the formatting seems to be a bit funky, or is it just me?

helpermethod:
Hi, it only looks that way because GitHub breaks the rather long `curl` invocation.

If you look at
https://github.com/sdkman/sdkman-cli/blob/c3469265208820e725f275f32d0ae6f750bd8542/contrib/completion/bash/sdk
everything seems to be nicely indented.

marc0der (approved):

PR #988, opened 2021-09-30
Fix version directory not created while installing

### LINKED ISSUE #987: Bug: Java installation failing
**Bug report**
Installation of java is failing with error:
```
mv: target '/home/aakarshituppal/.sdkman/candidates/java/8.0.302-open' is not a directory
```
Tried for versions 8.0.302-open, 11.0.12-open, 8.302.08.1-amzn

**To reproduce**
```
❯ sdk install java 8.0.302-open

Found a previously downloaded java 8.0.302-open archive. Not downloading it again...

Installing: java 8.0.302-open
mv: target '/home/aakarshituppal/.sdkman/candidates/java/8.0.302-open' is not a directory
Done installing!
```

**System info**
- OS: Ubuntu 20.04
- Shell: 
  - zsh 5.8 (x86_64-ubuntu-linux-gnu)
  - GNU bash, version 5.0.17(1)-release
- The output of `sdk version`: SDKMAN 5.12.4
-->

aksh1618:
`sdk install` command does not create the final version directory, causing `mv` command to fail.
Fixes #987

### DIFF
--- a/src/main/bash/sdkman-install.sh
+++ b/src/main/bash/sdkman-install.sh
@@ -67,7 +67,7 @@ function __sdkman_install_candidate_version() {
 	__sdkman_download "$candidate" "$version" || return 1
 	__sdkman_echo_green "Installing: ${candidate} ${version}"
 
-	mkdir -p "${SDKMAN_CANDIDATES_DIR}/${candidate}"
+	mkdir -p "${SDKMAN_CANDIDATES_DIR}/${candidate}/${version}"
 
 	rm -rf "${SDKMAN_DIR}/tmp/out"
 	unzip -oq "${SDKMAN_DIR}/archives/${candidate}-${version}.zip" -d "${SDKMAN_DIR}/tmp/out"


src/main/bash/sdkman-install.sh
```diff
@@ -67,7 +67,7 @@ function __sdkman_install_candidate_version() {
 	__sdkman_download "$candidate" "$version" || return 1
 	__sdkman_echo_green "Installing: ${candidate} ${version}"
 
-	mkdir -p "${SDKMAN_CANDIDATES_DIR}/${candidate}"
+	mkdir -p "${SDKMAN_CANDIDATES_DIR}/${candidate}/${version}"
```
marc0der, line 70: I don't think this is a good idea. This statement is only here to ensure that the candidates folder exists. Creation of the version directory happens when we move the archive contents here from the temp folder.

aksh1618:
Not required as the issue was due to personal shell config

Issue #965, opened 2021-09-04
Bug: Missing trailing newline character in java home command


irakliy01:
**Bug report**
With `-n` flag echo command does not add newline character at the end. So using **home** command ends up like this: 
```
irakliy-osx:~ irakliy$ sdk home java 11.0.12-zulu
/Users/irakliy/.sdkman/candidates/java/11.0.12-zuluirakliy-osx:~ irakliy$
```

**To reproduce**
Download SDKMAN 5.12.3, download java 11.0.12-zulu and then run `sdk home java 11.0.12-zulu`

**System info**

**OS**: MacOS Big Sur (11.4)
**Bash version**: GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin20)
Copyright (C) 2007 Free Software Foundation, Inc.
**Zsh version**: zsh 5.8 (x86_64-apple-darwin20.0)
**SDK version**: SDKMAN 5.12.3

marc0der:
Hi @irakliy01, as explained in Slack, this command is intended to be used in scripts. The original contributor wanted something to locate the home of a particular candidate version, and pipe the result into another command. This would break if you added a newline character at the end.

PR #990, opened 2021-10-03
Add trailing newline character in java home command

irakliy01:
This PR fixes [965](https://github.com/sdkman/sdkman-cli/issues/965) bug.

Need to update config file loaded on sdk install


marc0der:
Since this would break scripted clients using this command, I'll be closing this PR.

A new native version of this command will be released soon, but it will also not print a new line.

PR #991, opened 2021-10-07
Extended detection of platform on Raspberry Pi

FDelporte:


marc0der (approved):
Thanks for this! Of course, we also need to do quite a bit in the backend to accommodate it.

marc0der (changes requested):
Need to fix broken test and add new arm variants to existing tests.

PR #998, opened 2021-10-15
Add tests for Bash programmable completion

helpermethod:
Because Bash completion breaks regularly I've added a few tests for the most critical completions for

* sdk
* sdk install
* sdk install java


helpermethod:
I will think about testing the zsh completions too. Maybe there is a way for both completions to share some completion-agnostic code which can then be tested 🤔.

PR #1002, opened 2021-10-19
Unify Bash's and zsh's programmable completion

helpermethod:
At this stage this is nothing more than a PoC.


contrib/completion/bash/sdk
```diff
@@ -25,21 +25,19 @@ __sdkman_complete_command() {
...
 			done
 			;;
-		install|list)
-			local -r all_candidates=$(curl --silent "${SDKMAN_CANDIDATES_API}/candidates/all")
-
+		install|i|list|ls)
 			while IFS= read -d, -r candidate; do
 				candidates+=($candidate)
-			done <<< "$all_candidates"
+			done <<< "$(curl --silent "${SDKMAN_CANDIDATES_API}/candidates/all")"
```
marc0der, line 38: As in your question in Slack, it is better to use the cached candidates. Not only is it more performant, but it mitigates load on the candidates API.

contrib/completion/bash/sdk
```diff
@@ -73,12 +71,10 @@ __sdkman_complete_candidate_version() {
...
 			done
 			;;
-		install)
-			local -r all_candidate_versions=$(curl --silent "${SDKMAN_CANDIDATES_API}/candidates/$candidate/${SDKMAN_PLATFORM}/versions/all")
-
+		install|i)
 			while IFS= read -d, -r version; do
 				candidates+=($version)
-			done <<< "$all_candidate_versions"
+			done <<< "$(curl --silent "${SDKMAN_CANDIDATES_API}/candidates/$candidate/${SDKMAN_PLATFORM}/versions/all")"
```
marc0der, line 77: This is fine for now until we begin caching versions too. That will happen when the upgrade command is migrated to rust.

Issue #1005, opened 2021-10-20
Bug: Reinstalation of any cached archive fails


marc0der:
**Bug report**

Reinstallation of a previously installed version of Java fails. This always occurs when a cached artefact is already found in the `archives` folder.

**To reproduce**

```
❯ sdk install scala 3.1.0

...

❯ sdk rm scala 3.1.0

Deselecting scala 3.1.0...

Uninstalling scala 3.1.0...
 
❯ sdk install scala       

Found a previously downloaded scala 3.1.0 archive. Not downloading it again...
__sdkman_checksum_zip:15: no such file or directory: 

```

**System info**

Manjaro Linux
SDKMAN latest+4bf5b70
zsh 5.8 (x86_64-ubuntu-linux-gnu)

MLNW:
I'm experience the same issue.

marc0der:
@hgeraldino is currently working on this feature, so perhaps he can take a look.

hgeraldino:
Yes, I'll submit a patch to fix this bug shortly...

hgeraldino:
This has been addressed, I'll close this issue.

PR #1012, opened 2021-11-01
Fix bug when reinstalling a candidate (checksums for cached archives)

hgeraldino:
Issue https://github.com/sdkman/sdkman-cli/issues/1005

### DIFF
--- a/src/main/bash/sdkman-install.sh
+++ b/src/main/bash/sdkman-install.sh
@@ -127,8 +127,8 @@ function __sdkman_download() {
 	if [ ! -f "${archives_folder}/${candidate}-${version}.zip" ]; then
 		local platform_parameter="$(echo $SDKMAN_PLATFORM | tr '[:upper:]' '[:lower:]')"
 		local download_url="${SDKMAN_CANDIDATES_API}/broker/download/${candidate}/${version}/${platform_parameter}"
-		local zip_archive_target="${SDKMAN_DIR}/archives/${candidate}-${version}.zip"
 		local base_name="${candidate}-${version}"
+		local zip_archive_target="${SDKMAN_DIR}/archives/${base_name}.zip"
 		headers_file="${metadata_folder}/${base_name}.headers"
 
 		# pre-installation hook: implements function __sdkman_pre_installation_hook
--- a/src/main/bash/sdkman-install.sh
+++ b/src/main/bash/sdkman-install.sh
@@ -115,6 +115,7 @@ function __sdkman_install_local_version() {
 
 function __sdkman_download() {
 	local candidate version archives_folder
+	local headers_file
 
 	candidate="$1"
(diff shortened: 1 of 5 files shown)


src/main/bash/sdkman-install.sh
```diff
@@ -193,6 +194,12 @@ function __sdkman_checksum_zip() {
 	local algorithm checksum cmd
 	local shasum_avail=false
 	local md5sum_avail=false
+	
+	if [ ! -f "${headers_archive}" ]; then
```
marc0der, line 198: I must have overlooked this variable name in the past, but it's a bit confusing to me. It isn't pointing to a headers archive but a headers file. Would you mind renaming it for the sake of clarity?

marc0der (commented):
This looks good so far, thanks!

One scenario that might have been overlooked is the following:
Many JDKs are packaged as tarballs. So what we do (not to break caching) is repackage them as zips. When installing a previously installed JDK, we go looking for the zip file in the archives folder first.

I haven't verified this, but I suspect that the hash (for the tarball) will be used to check the repacked zip archive and will most probably fail.

The fact that we're repackaging this a as zip file isn't good, and I hope to improve on this in the future, but for now, it is what it is :smile:

src/main/bash/sdkman-install.sh
hgeraldino: s/headers*/headers_file//

hgeraldino:
I think we can handle those cases on the broker side, maybe by not writing the checksum headers for tarballs (so checks aren't performed by the `cli`), adding a `skip-checksum` header or something in that line.

Now I'm curious to know how repackaging these as zip files help with caching 🤔

marc0der:
So I wouldn't recommend we drop the checksums for tarballs, as most JDKs are packaged as such.

The JDKs were the first tarballs we brought onto the platform. Up to that point, they were all zips. When I introduced these tarballs, I didn't feel like rewriting the entire caching mechanism (which was geared for zips), so I decided to repackage the tarballs as zips in the interim. It has served its purpose well until now that we want to do some validation on the cached artefacts.

marc0der:
Also, I should have mentioned that we will revisit the caching strategy (if it's still needed because bandwidth is cheaper than when I initially wrote SDKMAN!) as we rewrite the installation process with Rust.

helpermethod:
IMHO the current caching strategy doesn't make sense anymore. It *very* rarely happens that I reinstall an SDK, and even then my crappy internet connection is good enough to download it on demand.

In the other hand all those cached archives occupied more than 4GB on my hard drive 🥲.

hgeraldino:
Ok, while the caching strategy is decided, this patch will just skip checksums for cached artifacts

marc0der (approved):
Awesome work, thank you!

PR #1020, opened 2021-12-03
"sdkman_checksum_skip" feature flag

hgeraldino:
Add feature flag to disable checksumming

### DIFF
--- a/src/main/bash/sdkman-install.sh
+++ b/src/main/bash/sdkman-install.sh
@@ -204,7 +204,13 @@ function __sdkman_checksum_zip() {
 		__sdkman_echo_yellow "Metadata file not found at '${headers_file}', skipping checksum..."
 		return
 	fi
-
+	
+	if [[ "$sdkman_checksum_enable" != "true" ]]; then
+		echo ""
+		__sdkman_echo_yellow "Checksums are disabled, skipping verification..."
+		return
+	fi
+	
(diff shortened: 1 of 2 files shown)


src/main/bash/sdkman-install.sh
```diff
@@ -204,6 +204,12 @@ function __sdkman_checksum_zip() {
 		__sdkman_echo_yellow "Metadata file not found at '${headers_file}', skipping checksum..."
 		return
 	fi
+	
+	if [[ "$sdkman_checksum_skip" == 'true' ]]; then
```
marc0der, line 208: I prefer to flip the config around to the positive to align with our other configurations. How about `sdkman_checksum_enable`. I would expect the default to be `false` for the time being until a bunch of us thoroughly test this.

src/main/bash/sdkman-install.sh
```diff
@@ -204,6 +204,12 @@ function __sdkman_checksum_zip() {
...
+		echo ""
+		__sdkman_echo_green "Checksums are disabled, skipping verification..."
```
marc0der, line 210: This is pure informational output, so it should be printed in yellow. We use green for a successful outcome and red for a failed outcome.

marc0der (approved):

marc0der:
Sorry, forgot to merge this one!

PR #1032, opened 2022-01-05
Feature/836

FDelporte:
Correct Linux version for ARMv6


marc0der:
The state of this PR isn't great with lots of merges from master etc. This PR has 8 commits for a single character that needs changing. Hope you don't mind but I'll cherry-pick the single commit that makes the change without merging your feature branch.

Just a suggestion, but in future consider rebasing your branch instead of merging from master branch.

marc0der:
I've cherry-picked only the two relevant commits and this is now deployed to the beta channel for testing. Closing this PR for now.

PR #1042, opened 2022-02-01
Add Homebrew support

helpermethod:
- [x]  add JReleaser Templates for the Homebrew Tap Formula and README
- [x] enable Homebrew packager in JReleaser's configuration
- [x] create a Personal Access Token with repo permissions to use as `JRELEASER_HOMEBREW_GITHUB_TOKEN`
- [x] add `sdkman_selfupdate_feature` flag to disable `selfupdate` functionality (including `auto_update`)
- [x] keep `sdkman_selfupdate_enable` flag for backwards-compatibility
- [x] introduce `sdkman_auto_update` flag for enabling/disabling `auto_update` (replaces `sdkman_selfupdate_enable`)
- [x] add tests to make sure that any combination of old / new flags works as expected


aalmiray:
Watch out for https://github.com/jreleaser/jreleaser/issues/702 when upgrading to JReleaser `v1.0.0-M2` or later. Templates will use `{{distributionExecutableUnix}}` instead of `{{distributionExecutable}}`.

helpermethod:
@marc0der Removing the `sdkman-selfupdate.sh` file as a post-install step is probably a little bit extreme but simplifies certain things. I'm still not sure if this is the best way 😆.

Alternatively we could introduce another flag for disabling the `selfupdate` functionality, and hope that no one switches it on.

helpermethod:
Hi @marc0der, I've added some tests to make sure that the old behaviour doesn't break.

src/main/bash/sdkman-init.sh
```diff
@@ -156,6 +156,8 @@ if [[ -z "${sdkman_curl_retry_max_time}" ]]; then sdkman_curl_retry_max_time=60;
 # set curl to continue downloading automatically
 if [[ -z "${sdkman_curl_continue}" ]]; then sdkman_curl_continue=true; fi
 
+if [[ -z "$sdkman_selfupdate_feature" ]]; then sdkman_selfupdate_feature=true; fi
```
marc0der, line 159: This should never happen if the config is added by the sdkman-hooks service in `install.sh` and `update.sh`. I think we can remove it.

src/test/groovy/sdkman/specs/SelfupdateSpec.groovy
```diff
@@ -0,0 +1,93 @@
...
+
+		then:
+		verifyOutput(bash.output)
+
+		where:
+		selfupdateFeature | selfupdateEnabled | autoUpdateEnabled | verifyOutput
+		"true"            | "true"            | "true"            | { it.contains("ATTENTION: A new version of SDKMAN is available...") }
+		"true"            | "true"            | "false"           | { it.contains("ATTENTION: A new version of SDKMAN is available...") }
+		"true"            | "false"           | "true"            | { it.contains("ATTENTION: A new version of SDKMAN is available...") }
+		"false"           | "true"            | "true"            | { !it.contains("ATTENTION: A new version of SDKMAN is available...") }
```
marc0der, line 91: Awesome, I sometimes forget how great Spock is. Nice work!

marc0der (commented):
Looks good so far. We shouldn't forget to add the new config to the sdkman-hooks service as mentioned.

src/main/bash/sdkman-init.sh
helpermethod: I thought about not exposing the flag to non-Homebrew users, but I can see that it makes sense to add it to the sdkman-hooks service.

marc0der: Perhaps you can add a comment before to say why you've got this line here.

helpermethod:
@marc0der With the hook updates promoted to stable, I wonder if

```
# Attempt upgrade after all is done
if [[ "$COMMAND" != "selfupdate" && "$sdkman_selfupdate_feature" == "true" && ("$sdkman_auto_update" == "true" || "$sdkman_selfupdate_enable" == "true") ]]; then
	__sdkman_auto_update "$SDKMAN_REMOTE_VERSION" "$SDKMAN_VERSION"
fi
```

the `sdkman_selfupdate_enable` check here is actually needed? I guess the selfupdate hook will take care of replacing it, right?

Would be nice if we could get rid of that OR expression.

marc0der:
@helpermethod You are right. The selfupdate should take care of it, provided we deploy the hooks before the CLI. That said, I suggest leaving it in until both have been deployed, then removing it in a tiny follow-up PR once the dust has settled.

helpermethod:
Too late, already removed it 😂

marc0der:
@helpermethod You are too fast :laughing: You probably need to update the `SdkmanBashEnvBuilder` and `SelfupdateSpec` too which still refer to the old variable.

helpermethod:
Removed the `selfupdate_enable` flag from the data table.

helpermethod:
Oh, and before we finally do the release I need to drop the old SDKMAN! Homebrew repository.

marc0der:
Okay, are you happy for me to merge that so we can test it in the beta channel?

helpermethod:
Let's go!

src/test/groovy/sdkman/specs/SelfupdateSpec.groovy
```diff
@@ -0,0 +1,91 @@
...
+		bash.execute("source $bootstrapScript")
+
+		when:
+		bash.execute("sdk version")
+
+		then:
+		verifyOutput(bash.output)
+
+		where:
+		selfupdateFeature | autoUpdateEnabled | verifyOutput
```
marc0der, line 86: Could we also test the negative cases here? It would be good to assert when `!it.contains("ATTENTION...")`.

helpermethod: I think there is at least one test for the negative case, no?

marc0der: Ah sorry, missed that. All good then!

felipecrs:
This was a breaking change for me, sad it wasn't tagged as one. I think it should be avoided to perform breaking changes without bumping the major version (like from 5 to 6).

https://github.com/felipecrs/dotfiles/commit/5ba2f2b338282335b27969ebaf4c033aeabe4125

helpermethod:
Sorry to hear 😥!

The `sdk selfupdate` command will automatically perform the migration, so for most users it should be a backwards-compatible change.

That's why we didn't bump the major version.

felipecrs:
Got it. Well, https://github.com/sdkman/sdkman-cli/pull/1069 kind of prevents it (in the sense that if someone didn't touch the default configuration, it will be handled by the migration). But only in that case.

felipecrs:
By the way, docs are missing the new `sdkman_auto_update`:

https://sdkman.io/usage#config

PR #1054, opened 2022-02-20
Sanitise noise from metadata headers file

marc0der:
On the download of a binary, the curl headers dump is now written to a temporary file. In retrospect, we sanitise this before writing it to the `~/.sdkman/var/metadata` folder. Only lines starting with `X-Sdkman` will now be recorded.

### DIFF
--- a/src/main/bash/sdkman-install.sh
+++ b/src/main/bash/sdkman-install.sh
@@ -115,7 +115,7 @@ function __sdkman_install_local_version() {
 
 function __sdkman_download() {
 	local candidate version archives_folder
-	local headers_file
+	local headers_file tmp_headers_file
 
 	candidate="$1"
 	version="$2"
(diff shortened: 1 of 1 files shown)


src/main/bash/sdkman-install.sh
```diff
@@ -150,7 +151,8 @@ function __sdkman_download() {
 		echo ""
 
 		# download binary
-		__sdkman_secure_curl_download "${download_url}" --output "${binary_input}" --dump-header "${headers_file}"
+		__sdkman_secure_curl_download "${download_url}" --output "${binary_input}" --dump-header "${tmp_headers_file}"
+		grep '^X-Sdkman' "${tmp_headers_file}" > "${headers_file}"
```
hgeraldino, line 155: Don’t recall if the message key is spelled SdkMan or Sdkman, maybe do grep -i to ignore case?

marc0der: I checked in the broker service and subsequently tested it with Gradle 2.11 that you uploaded with a checksum. Works perfectly. Also, all our checksum validation tests are still green.

hgeraldino: Should we delete the temp file as well?

marc0der: I thought I'd leave it for `flush` to clean up. That way we can go back and trace if something goes wrong.

hgeraldino (approved):

PR #1063, opened 2022-03-09
Fix selfupdate on stable and beta channels

marc0der:


src/main/bash/sdkman-selfupdate.sh
```diff
@@ -57,7 +58,7 @@ function __sdkman_auto_update() {
 		fi
 
 		if [[ "$upgrade" == "Y" || "$upgrade" == "y" ]]; then
-			__sdk_selfupdate
+			__sdk_selfupdate "force"
```
helpermethod, line 61: Why is the `force` needed here?

marc0der: Because shellcheck complains if there is no parameter applied to that function (we reference `$1` inside the function). The only value it can be is `force`.

helpermethod (approved):
LGTM!

PR #1069, opened 2022-03-14
Allow `~/.config/sdkman/config` to override `~/.sdkman/etc/config`

### LINKED ISSUE #941: Feature: stop mixing the default config with user config in a single file
- [X] I have read the [CONTRIBUTING guidelines](CONTRIBUTING.md)

**Feature request**

All started with https://github.com/sdkman/sdkman-cli/pull/928, which happened after I decided to include sdkman's config in my dotfiles repository: https://github.com/felipecrs/dotfiles/blob/c83fc79656129c35aa39501bab1c58f4a5bd683a/dot_sdkman/etc/config

So it turns out that sdkman expects some user configuration variable to be set, and treat them as the default settings. But it may not be a good idea to change the user's config file unattendedly during a self-update or on installation.

More insights comes from https://github.com/sdkman/sdkman-cli/pull/927#discussion_r664130904:

> I'm 100% against assuming everyone will have this set in their config files. For example, I manage my dotfiles using another 
> program, so I'm the one who makes changes to it. If later in the future sdkman introduces new config variables, I would miss it as my dotfiles manager would override it with my own wrote config again.
> 
> I think there are 2 different things that should not be mixed:
> 
> Default settings
> User preferences (the `~/.sdkman/etc/config` file)
> And I don't think sdkman should make changes unattended(ly) to the user preferences, but only to the default settings. Just like Git (`~/.gitconfig`) or any other tool.

So my proposal is:

Have the `~/.sdkman/etc/config` as the default configuration source. This file can be included in this repository, and it does not need to be primed during an installation or dynamically updated during a self-update. An update of SDKMAN will contain the file, just like the other files.

Inside of this file, all the possible configuration variables will be set with their default values (some of them can be removed from `sdkman-init.sh` thus, such as the `curl` related ones).

Then, an user can optionally set its own configuration options in another file. And by following XDG's recommendation, this file should be at `~/.config/sdkman/config`/`~/.config/sdkman/rc`. Or `~/.sdkmanconfig`/`~/.sdkmanrc`.

The sdkman-main should read the configuration this way:

```bash
# load default settings
source $SDKMAN_DIR/etc/config

# load user settings
source $XDG_CONFIG_HOME/sdkman/config
```

The latter instruction will override any variables set with the default settings.

felipecrs:
Because the current approach is very dotfiles unfriendly.

Fixes #941

### DIFF
--- a/src/main/bash/sdkman-config.sh
+++ b/src/main/bash/sdkman-config.sh
@@ -26,5 +26,5 @@ function __sdk_config() {
 		return 1
 	fi
 
-	"${editor[@]}" "${SDKMAN_DIR}/.sdkmanrc"
+	"${editor[@]}" "${SDKMAN_DIR}/etc/config"
 }
--- a/src/main/bash/sdkman-config.sh
+++ b/src/main/bash/sdkman-config.sh
@@ -26,5 +26,5 @@ function __sdk_config() {
 		return 1
 	fi
 
-	"${editor[@]}" "${SDKMAN_DIR}/etc/config"
+	"${editor[@]}" "${SDKMAN_DIR}/.sdkmanrc"
 }
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -34,6 +34,11 @@ if [ -f "${SDKMAN_DIR}/etc/config" ]; then
 	source "${SDKMAN_DIR}/etc/config"
 fi
 
+# Load the sdkman user config if it exists.
+if [ -f "${HOME}/.sdkmanrc" ]; then
+	source "${SDKMAN_DIR}/.sdkmanrc"
+fi
+
 # infer platform
 function infer_platform() {
 	local kernel
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -35,8 +35,8 @@ if [ -f "${SDKMAN_DIR}/etc/config" ]; then
 fi
 
 # Load the sdkman user config if it exists.
-if [ -f "${HOME}/.sdkmanrc" ]; then
-	source "${SDKMAN_DIR}/.sdkmanrc"
+if [ -f "${HOME}/.sdkmanconfig" ]; then
+	source "${HOME}/.sdkmanconfig"
 fi
 
 # infer platform
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -35,8 +35,8 @@ if [ -f "${SDKMAN_DIR}/etc/config" ]; then
 fi
 
 # Load the sdkman user config if it exists.
-if [ -f "${HOME}/.sdkmanconfig" ]; then
-	source "${HOME}/.sdkmanconfig"
+if [ -f "${HOME}/.config/sdkman/config" ]; then
+	source "${HOME}/.config/sdkman/config"
 fi
 
 # infer platform
--- a/src/main/bash/sdkman-main.sh
+++ b/src/main/bash/sdkman-main.sh
@@ -96,8 +96,8 @@ function sdk() {
 	fi
 
 	# Load the sdkman user config if it exists.
-	if [ -f "${HOME}/.sdkmanrc" ]; then
-		source "${SDKMAN_DIR}/.sdkmanrc"
+	if [ -f "${HOME}/.sdkmanconfig" ]; then
+		source "${HOME}/.sdkmanconfig"
 	fi
 
 	# no command provided
--- a/src/main/bash/sdkman-main.sh
+++ b/src/main/bash/sdkman-main.sh
@@ -95,6 +95,11 @@ function sdk() {
 		source "${SDKMAN_DIR}/etc/config"
 	fi
 
+	# Load the sdkman user config if it exists.
+	if [ -f "${HOME}/.sdkmanrc" ]; then
+		source "${SDKMAN_DIR}/.sdkmanrc"
+	fi
+
 	# no command provided
 	if [[ -z "$COMMAND" ]]; then
 		__sdk_help
--- a/src/main/bash/sdkman-main.sh
+++ b/src/main/bash/sdkman-main.sh
@@ -96,8 +96,8 @@ function sdk() {
 	fi
 
 	# Load the sdkman user config if it exists.
-	if [ -f "${HOME}/.sdkmanconfig" ]; then
-		source "${HOME}/.sdkmanconfig"
+	if [ -f "${HOME}/.config/sdkman/config" ]; then
+		source "${HOME}/.config/sdkman/config"
 	fi
 
 	# no command provided


felipecrs:
I may need to add tests and fix the CI. I would just like to receive a greenlight from the maintainers before I spend time on it.

marc0der (commented):
I'd really prefer to keep the base SDKMAN directory free from configuration. That is explicitly what the etc directory is for. Also, you will see no other files, just directories, in the base dir.

However, I would consider an override if you wanted an `.sdkmanrc` file in your home directory. That makes more sense and is more in line with what you would expect on a Unix system.

Another thing also not considered is the `sdk config` command for editing the configuration. Which of these files should it be responsible for editing?

felipecrs:
> However, I would consider an override if you wanted an .sdkmanrc file in your home directory. That makes more sense and is more in line with what you would expect on a Unix system.

That's what I'm proposing, `~/.sdkmanrc` as an override option only. I have no intention to change anything in `~/.sdkman/etc/config` or how it works.

> Another thing also not considered is the sdk config command for editing the configuration. Which of these files should it be responsible for editing?

Actually, I considered and changed `sdk config` to edit `~/.sdkmanrc` instead of `~/.sdkman/etc/config`. But I'm okay to change it back to `etc/config` if you do not agree.

helpermethod:
In any case please don't name it `.sdkmanrc` 😂. We already use `.sdkmanrc` files for storing per-project candidate versions.

An `.sdkmanrc` file in the $HOME directory looks like a similar file but per-user.

A lot of popular tools use that convention, e.g. NPM and direnv

https://docs.npmjs.com/cli/v8/configuring-npm/npmrc
https://direnv.net/#how-it-works

src/main/bash/sdkman-config.sh
```diff
@@ -26,5 +26,5 @@ function __sdk_config() {
 		return 1
 	fi
 
-	"${editor[@]}" "${SDKMAN_DIR}/etc/config"
+	"${editor[@]}" "${SDKMAN_DIR}/.sdkmanrc"
```
marc0der, line 29: This is pointing to `${SDKMAN_DIR}/.sdkmanrc`, not `${HOME}/.sdkmanrc`

src/main/bash/sdkman-init.sh
```diff
@@ -34,6 +34,11 @@ if [ -f "${SDKMAN_DIR}/etc/config" ]; then
 	source "${SDKMAN_DIR}/etc/config"
 fi
 
+# Load the sdkman user config if it exists.
+if [ -f "${HOME}/.sdkmanrc" ]; then
+	source "${SDKMAN_DIR}/.sdkmanrc"
```
marc0der, line 39: Same as above. You are sourcing a file inside the `SDKMAN_DIR` folder.

src/main/bash/sdkman-main.sh
```diff
@@ -95,6 +95,11 @@ function sdk() {
 		source "${SDKMAN_DIR}/etc/config"
 	fi
 
+	# Load the sdkman user config if it exists.
+	if [ -f "${HOME}/.sdkmanrc" ]; then
+		source "${SDKMAN_DIR}/.sdkmanrc"
```
marc0der, line 100: And again.

src/main/bash/sdkman-config.sh
felipecrs: Ops, my bad!

felipecrs:
How about `~/.sdkmanconfig` (like `~/.gitconfig`)?

src/main/bash/sdkman-config.sh
marc0der: We need to think carefully about which file has precedence when editing. I don't think `$HOME/.sdkmanrc` should take precedence.

felipecrs: Changed back to etc/config then.

tripleo1:
I would prefer not to have another file in my home directory.

Use .config or something.

Just my 2 cents.

felipecrs:
Yeah, I also agree.

felipecrs:
Changed.

marc0der:
I'll be very honest, but I'm leaning towards not having this capability at all. I honestly don't think that our users want this feature since it has never been asked for before now.

Another point is that modern Linux systems do not store configuration directly in the home directory. Instead, the `~/config` directory is used to contain all config files. And again different rules apply on Mac.

Besides that, I think that this has many undesired side effects and consequences that simply haven't been thought through.

felipecrs:
@marc0der I changed already to ~/.config.

patrikcoch123 (approved):

Li-Gru (approved):

felipecrs:
> I honestly don't think that our users want this feature since it has never been asked for before.

@marc0der it seems that I'm no longer alone: https://github.com/sdkman/sdkman-cli/issues/1160

If you want, I can rebase this PR.

untainsYD:
Are there any updates on this issue? It is quite annoying that `sdk-cli` pollutes my home directory instead of creating `.sdkmanrc` file in `~/.config/sdkman/...`

felipecrs:
I will wait for maintainers to comment before I rebase this PR.

src/main/bash/sdkman-init.sh
```diff
@@ -34,6 +34,11 @@ if [ -f "${SDKMAN_DIR}/etc/config" ]; then
 	source "${SDKMAN_DIR}/etc/config"
 fi
 
+# Load the sdkman user config if it exists.
+if [ -f "${HOME}/.config/sdkman/config" ]; then
```
sdavids, line 38: `"${XDG_CONFIG_HOME}/sdkman/config"`

if `XDG_CONFIG_HOME` is not set then `"${HOME}/.config/sdkman/config"`

https://wiki.archlinux.org/title/XDG_Base_Directory#User_directories

src/main/bash/sdkman-main.sh
```diff
@@ -95,6 +95,11 @@ function sdk() {
 		source "${SDKMAN_DIR}/etc/config"
 	fi
 
+	# Load the sdkman user config if it exists.
+	if [ -f "${HOME}/.config/sdkman/config" ]; then
```
sdavids, line 99: see above

sdavids:
> Are there any updates on this issue? It is quite annoying that `sdk-cli` pollutes my home directory instead of creating `.sdkmanrc` file in `~/.config/sdkman/...`

```shell
$ export SDKMAN_DIR="${XDG_DATA_HOME}/sdkman" && \
  curl -s "https://get.sdkman.io/" | bash
```

~/.zshrc
```shell
### sdkman - https://sdkman.io/

readonly sdkman_home="${XDG_DATA_HOME}/sdkman"

if [ -d "${sdkman_home}" ]; then
  export SDKMAN_DIR="${sdkman_home}"

  if [ -s "${sdkman_home}/bin/sdkman-init.sh" ]; then
    . "${sdkman_home}/bin/sdkman-init.sh"
  fi
fi
```

It would be nice if there were a `SDKMAN_CONFIG_DIR`:

```shell
$ export SDKMAN_DIR="${XDG_DATA_HOME}/sdkman" && \
  export SDKMAN_CONFIG_DIR="${XDG_CONFIG_HOME}/sdkman" && \
  curl -s "https://get.sdkman.io/" | bash
```

felipecrs:
@sdavids thanks for your suggestions, but I'll wait to have a go ahead from some maintainer before spending more time in this PR.

sdavids:
> It would be nice if there were a `SDKMAN_CONFIG_DIR`:
> 
> ```shell
> $ export SDKMAN_DIR="${XDG_DATA_HOME}/sdkman" && \
>   export SDKMAN_CONFIG_DIR="${XDG_CONFIG_HOME}/sdkman" && \
>   curl -s "https://get.sdkman.io/" | bash
> ```

@marc0der

felipecrs:
I no longer use SDKMAN!, and I don't think there's any intention to merge this PR anyway, so I'm closing it.

If some maintainer wants to merge it, please let me know and I can reopen and fix its conflicts.

PR #1077, opened 2022-03-25
Remove local caching of candidates

hgeraldino:
Don't copy the `${zip_output}` file to `${SDKMAN_DIR}/candidates`; leave it on the `tmp` dir and explode it from there.

I'm thinking of adding yet one more `rm -r tmp/*` after the install is done, but will submit as a separate PR once we deal with the removal of tarball->zip conversions


helpermethod (approved):
Great MR! You thought of everything, even the completions and the Homebrew Formula
💪 !

src/main/bash/sdkman-install.sh
```diff
@@ -114,65 +114,56 @@ function __sdkman_install_local_version() {
...
+	local pre_installation_hook="${SDKMAN_DIR}/tmp/hook_pre_${candidate}_${version}.sh"
+	__sdkman_echo_debug "Get pre-installation hook: ${SDKMAN_CANDIDATES_API}/hooks/pre/${candidate}/${version}/${platform_parameter}"
+	__sdkman_secure_curl "${SDKMAN_CANDIDATES_API}/hooks/pre/${candidate}/${version}/${platform_parameter}" >| "$pre_installation_hook"
+	__sdkman_echo_debug "Copy remote pre-installation hook: $pre_installation_hook"
+	source "$pre_installation_hook"
+	__sdkman_pre_installation_hook || return 1
+	__sdkman_echo_debug "Completed pre-installation hook..."
+
+	export local binary_input="${SDKMAN_DIR}/tmp/${base_name}.bin"
+	export local zip_output="${SDKMAN_DIR}/tmp/${base_name}.zip"
```
marc0der, line 141: The target file here is a zip, but only because the post-install hook transforms the tarball into a zip. The hook will need to determine if it is a tarball or zip based on the content header, then place it here either as `.tar.gz` or `.zip`. Then we will need to deal with it appropriately.

src/main/bash/sdkman-install.sh
```diff
@@ -70,7 +70,7 @@ function __sdkman_install_candidate_version() {
 	mkdir -p "${SDKMAN_CANDIDATES_DIR}/${candidate}"
 
 	rm -rf "${SDKMAN_DIR}/tmp/out"
-	unzip -oq "${SDKMAN_DIR}/archives/${candidate}-${version}.zip" -d "${SDKMAN_DIR}/tmp/out"
+	unzip -oq "${SDKMAN_DIR}/tmp/${candidate}-${version}.zip" -d "${SDKMAN_DIR}/tmp/out"
```
marc0der, line 73: This is a danger zone :skull:
All binaries will be placed here, including tarballs. As the PR stands now, I think it will break for all tarball distributions. That means the majority of the Java SDKs too.

src/main/bash/sdkman-install.sh
```diff
@@ -114,65 +114,56 @@ function __sdkman_install_local_version() {
...
-	fi
-	__sdkman_validate_zip "${archives_folder}/${candidate}-${version}.zip" || return 1
-	__sdkman_checksum_zip "${archives_folder}/${candidate}-${version}.zip" "${headers_file}" || return 1
+	# download binary
+	__sdkman_secure_curl_download "${download_url}" --output "${binary_input}" --dump-header "${tmp_headers_file}"
+	grep '^X-Sdkman' "${tmp_headers_file}" > "${headers_file}"
+	__sdkman_echo_debug "Downloaded binary to: ${binary_input} (HTTP headers written to: ${headers_file})"
+
+	# post-installation hook: implements function __sdkman_post_installation_hook
+	# responsible for taking `binary_input` and producing `zip_output`
```
marc0der, line 155: This is what I was describing earlier...

marc0der (changes requested):
Everything looks great, but we need to find a solution for dealing with the types of binary content that we download to the `.bin` file. We could probably do this work inside the post-install hooks. Apart from that, it looks fantastic!

marc0der (approved):

Issue #1087, opened 2022-04-16
Bug: Conflict with hstr application when set sdkman_auto_env=true


speedlog:
**Bug report**
When i have installed hstr (https://github.com/dvorka/hstr) and sdkman with sdkman_auto_env=true

Every command i get message "bash: PROMPT_COMMAND: line X: `history -a; history -n; ;sdkman_auto_env'"

```
hstr version "2.3.0" (2020-11-19T07:41:00)

bash: PROMPT_COMMAND: line 0: syntax error near unexpected token `;'
bash: PROMPT_COMMAND: line 0: `history -a; history -n; ;sdkman_auto_env'

user:~/Downloads/test-sdk-env/car-app$ java -version
java version "17.0.1" 2021-10-19 LTS
Java(TM) SE Runtime Environment (build 17.0.1+12-LTS-39)
Java HotSpot(TM) 64-Bit Server VM (build 17.0.1+12-LTS-39, mixed mode, sharing)
bash: PROMPT_COMMAND: line 1: syntax error near unexpected token `;'
bash: PROMPT_COMMAND: line 1: `history -a; history -n; ;sdkman_auto_env'

user:~/Downloads/test-sdk-env/car-app$ cd ..
bash: PROMPT_COMMAND: line 2: syntax error near unexpected token `;'
bash: PROMPT_COMMAND: line 2: `history -a; history -n; ;sdkman_auto_env'
```

There is a resolution.
Despite of warning comment ("THIS MUST BE AT THE END OF THE FILE FOR SDKMAN TO WORK!!!") i need to set hstr fragment in `.bashrc` after sdkman fragment.
Something like that:

```
#THIS MUST BE AT THE END OF THE FILE FOR SDKMAN TO WORK!!!
export SDKMAN_DIR="/home/user/.sdkman"
[[ -s "/home/user/.sdkman/bin/sdkman-init.sh" ]] && source "/home/user/.sdkman/bin/sdkman-init.sh"

# HSTR configuration - add this to ~/.bashrc
alias hh=hstr                    # hh to be alias for hstr
export HSTR_CONFIG=hicolor       # get more colors
shopt -s histappend              # append new history items to .bash_history
export HISTCONTROL=ignorespace   # leading space hides commands from history
export HISTFILESIZE=10000        # increase history file size (default is 500)
export HISTSIZE=${HISTFILESIZE}  # increase history size (default is 500)
# ensure synchronization between bash memory and history file
[...]

PR #1088, opened 2022-04-16
#1087 Fix sdkman_auto_env when PROMPT_COMMAND ends with semicolon

speedlog:
Remove double semicolon problem
Without this fix there may be a message with double semicolon:
bash: PROMPT_COMMAND: line 0: `history -a; history -n; ;sdkman_auto_env'

### DIFF
--- a/src/main/bash/sdkman-init.sh
+++ b/src/main/bash/sdkman-init.sh
@@ -225,8 +225,9 @@ if [[ "$sdkman_auto_env" == "true" ]]; then
 
 			export SDKMAN_OLD_PWD="$PWD"
 		}
-
-		[[ -z "$PROMPT_COMMAND" ]] && PROMPT_COMMAND="sdkman_auto_env" || PROMPT_COMMAND="${PROMPT_COMMAND%\;};sdkman_auto_env"
+		
+		trimmed_prompt_command="${PROMPT_COMMAND%"${PROMPT_COMMAND##*[![:space:]]}"}"
+		[[ -z "$trimmed_prompt_command" ]] && PROMPT_COMMAND="sdkman_auto_env" || PROMPT_COMMAND="${trimmed_prompt_command%\;};sdkman_auto_env"
 	fi
 
 	sdkman_auto_env


src/main/bash/sdkman-init.sh
```diff
@@ -225,8 +225,9 @@ if [[ "$sdkman_auto_env" == "true" ]]; then
 
 			export SDKMAN_OLD_PWD="$PWD"
 		}
-
-		[[ -z "$PROMPT_COMMAND" ]] && PROMPT_COMMAND="sdkman_auto_env" || PROMPT_COMMAND="${PROMPT_COMMAND%\;};sdkman_auto_env"
+		
+		TRIMMED_PROMPT_COMMAND="${PROMPT_COMMAND%%*( )}"
```
helpermethod, line 229: We normally use all uppercase var names only for environment variables.

So something like `trimmed_prompt_command` would be more fitting.

helpermethod, line 229: Does this parameter expansion really work? I've tested it on a few example strings and it didn't remove trailing spaces.

speedlog: Ok, tommorow i will change.

speedlog: I've tested it using bash. Tomorrow I will check if extglob is on, but I have default config of Ubuntu.

I will also prepare some tests to easy run for example in docker.

Maybe I should use another method like sed/awk, but I wanted to have minimal dependencies. If you prefere this solution i would Change my commit.

Moment ago i've read about xargs but it removes  double spaces also in the middle of text so it does too many for our needs.

helpermethod: This should work:

```
trimmed_prompt_command="${PROMPT_COMMAND%"${PROMPT_COMMAND##*[![:space:]]}"}"
```

(taken straight from the Bash Bible: https://github.com/dylanaraps/pure-bash-bible)

speedlog: I used your fragment.
Default in Ubuntu extglob is on.

Prepare some test script, and it works:
```bash
#!/bin/bash
test_prompt () {
	test_prompt=$1
	trimmed_prompt_command="${test_prompt%"${test_prompt##*[![:space:]]}"}"
	[[ -z "$trimmed_prompt_command" ]] && test_prompt="sdkman_auto_env" || test_prompt="${trimmed_prompt_command%\;};sdkman_auto_env"
	echo $test_prompt
}

# expected output for each input: aaa;sdkman_auto_env
test_prompt "aaa"
test_prompt "aaa "
test_prompt "aaa;"
test_prompt "aaa; "
```

Of course i've tested change in commit with my local sdkman and installed hstr application (root cause of this issue).

helpermethod (approved):
LGTM!

helpermethod:
@speedlog Thanks for contributing! If you feel up to the task you can also add some more tests, the common case should already be covered.

speedlog:
@helpermethod i could add some tests :)
I've checked and see Spock and Cucumber tests.
I think it should be Spock test.
My concept is extend `SdkmanBashEnvBuilder` to allow set `PROMPT_COMMAND` in env and add some tests in `EnvCommandSpec`. Please confirm if it's right way or should i done it in another way.

helpermethod:
> @helpermethod i could add some tests :)
> I've checked and see Spock and Cucumber tests.
> I think it should be Spock test.
> My concept is extend `SdkmanBashEnvBuilder` to allow set `PROMPT_COMMAND` in env and add some tests in `EnvCommandSpec`. Please confirm if it's right way or should i done it in another way.

That is exactly the way to go 👍!

Weltraumschaf:
Is this reelased in 5.15.0? Since today my dotfiles setup on all machines is broken with this error 😞

Weltraumschaf:
> The last release was 3 months ago

After poking around in your repo I saw this. I updated my `PROMPT_COMMAND` [some time ago](https://github.com/Weltraumschaf/dotfiles/blob/master/src/dotfiles/_bashrc#L58). I have no idea why it broke days/weeks after that and not immediately 🤷‍♂️ 

My current workaround is:

```shell
if [ -n "${PROMPT_COMMAND}" ]; then
    export PROMPT_COMMAND="history -a; history -c; history -r; ${PROMPT_COMMAND}"
else 
    export PROMPT_COMMAND="history -a; history -c; history -r"
fi
```

PR #1089, opened 2022-04-22
Add new script to choose candidate interactively

jasondlee:
This script lists all installed versions for the specified candidate and allows the user to choose one by a number. Delegates to __sdk_use to set the candidate/version for the current shell.

### DIFF
--- a/src/main/bash/sdkman-choose.sh
+++ b/src/main/bash/sdkman-choose.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+
+#
+#   Copyright 2022 Jason Lee
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


helpermethod:
Hi @jasondlee! 

I like your idea, but there seems to be a lot of overlap with using the Bash / zsh autocompletion.

@marc0der wdyt?

marc0der:
I agree with @helpermethod. It is already provided mainly by autocomplete. Also, if we were to consider something like this, I would expect to see complete test coverage for a new feature like this.

jasondlee:
> I like your idea, but there seems to be a lot of overlap with using the Bash / zsh autocompletion.

Hmm. I guess I've missed something, then. I'll read up on the autocomplete. If that works, I'm happy. :)

jasondlee:
Yup. Tab complete works just fine (for use, but not for remove, fwiw). Not sure how I missed that. At least I learned a bit about the internals of sdkman. :)

helpermethod:
> Yup. Tab complete works just fine (for use, but not for remove, fwiw). Not sure how I missed that. At least I learned a bit about the internals of sdkman. :)

The completion works for

`sdk rm` or `sdk uninstall`

though 😁.

jasondlee:
Good to know. Thanks!

PR #1108, opened 2022-07-31
fix(completion): fix static analysis shellcheck errors

leagris:
In ./contrib/completion/bash/sdk line 36:
			candidates=${SDKMAN_CANDIDATES[@]}
                        ^--------^ SC2178 (warning): Variable was used as an array but is now assigned a string.
                                   ^---------------------^ SC2124 (warning): Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.

In ./contrib/completion/bash/sdk line 52:
	COMPREPLY=($(compgen -W "${candidates[*]}" -- "$current_word"))
                   ^-- SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).

In ./contrib/completion/bash/sdk line 79:
	COMPREPLY=($(compgen -W "${candidates[*]}" -- "$candidate_version"))
                   ^-- SC2207 (warning): Prefer mapfile or read -a to split command output (or quote to avoid splitting).

For more information:
  https://www.shellcheck.net/wiki/SC2124 -- Assigning an array to a string! A...
  https://www.shellcheck.net/wiki/SC2178 -- Variable was used as an array but...
  https://www.shellcheck.net/wiki/SC2207 -- Prefer mapfile or read -a to spli...

### DIFF
--- a/contrib/completion/bash/sdk
+++ b/contrib/completion/bash/sdk
@@ -33,7 +33,7 @@ __sdkman_complete_command() {
 			done
 			;;
 		install|i|list|ls)
-			candidates=${SDKMAN_CANDIDATES[@]}
+			candidates=("${SDKMAN_CANDIDATES[@]}")
 			;;
 		env|e)
 			candidates=("init" "install" "clear")
@@ -49,7 +49,7 @@ __sdkman_complete_command() {
 			;;
 	esac
 
-	COMPREPLY=($(compgen -W "${candidates[*]}" -- "$current_word"))
+	read -r -a COMPREPLY < <(compgen -W "${candidates[*]}" -- "$current_word")
 }
 
 __sdkman_complete_candidate_version() {
@@ -76,7 +76,7 @@ __sdkman_complete_candidate_version() {
 			;;
 	esac
 
-	COMPREPLY=($(compgen -W "${candidates[*]}" -- "$candidate_version"))
+	read -r -a COMPREPLY < <(compgen -W "${candidates[*]}" -- "$candidate_version")
 }
 
 complete -o default -F _sdk sdk


contrib/completion/bash/sdk
```diff
@@ -49,7 +49,7 @@ __sdkman_complete_command() {
 			;;
 	esac
 
-	read -r -a COMPREPLY < <(compgen -W "${candidates[*]}" -- "$current_word")
+	mapfile -t COMPREPLY < <(compgen -W "${candidates[*]}" -- "$current_word")
```
helpermethod, line 52: Hi, we are deliberately not using process substitution, as more often than not users run SDKMAN! in Bash POSIX mode, which among other things disables process substitution (that's probably why some tests fail, as they are ran with POSIX mode enabled).

contrib/completion/bash/sdk
```diff
@@ -76,7 +76,7 @@ __sdkman_complete_candidate_version() {
 			;;
 	esac
 
-	read -r -a COMPREPLY < <(compgen -W "${candidates[*]}" -- "$candidate_version")
+	mapfile -t COMPREPLY < <(compgen -W "${candidates[*]}" -- "$candidate_version")
```
helpermethod, line 79: `mapfile` requires Bash 4+, so this may introduce a breaking change for some of our users (even if using `mapfile` is technically better what using `read`).

helpermethod:
Hi @leagris,

I've added a few comments to your proposed changes. While these are technically correct, we highly value backwards compatibility, which some of these changes might break.

contrib/completion/bash/sdk
leagris: Thank-you for the info. It was not straightforward at all because Bash5.1 accepts the constructs in posix mode. I found alternative method to plase old Bash posix mode with mapfile array from command output.

contrib/completion/bash/sdk
leagris: The original implementation was using an horrible word splitting on `IFS` using unquoted variable expansion, which also expand globbing patterns within variable. I could manage to have `read -a` loads the newline delimited entries generated by the `compgen` command. That put it at the 15 years old obsolete Bash 3.2 compatibility. Now it will sure not work with Zsh, which use its own custom grammar for arrays.

As for the failing tests. I really can't make sense of it at all.

PR #1141, opened 2022-10-28
Fix test failures in OSX/Intel when inferring the platform

### LINKED ISSUE #1137: Bug: Tests fail on macs because of a bad platform string
**Bug report**

Many tests fail on macs because the string used in `UnixUtils.inferPlatform(String, String)` as a lookup key for `UnixUtils.platforms` was changed a while ago (in commit 06fd0436a6b9ac79ba2b28125637cfde2b3c9a3b) to be obtained from `uname -s` instead of from `System.getProperty("os.name")`; the value of `uname -s` for macs is `Darwin`, so the hardcoded mac platform  key must be correspondingly changed.

**To reproduce**

Run tests on a mac.

**System info**

macOS 12.6
zsh 5.9
SDKMAN 5.16.0

hgeraldino:
Fixes #1137

### DIFF
--- a/src/test/groovy/sdkman/support/UnixUtils.groovy
+++ b/src/test/groovy/sdkman/support/UnixUtils.groovy
@@ -8,6 +8,9 @@ class UnixUtils {
 			],
 			"Mac OS X": [
 					"x86_64": "DarwinX64",
+			],
+			"Darwin": [
+					"x86_64": "DarwinX64",
 			]
 	]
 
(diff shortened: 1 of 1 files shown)


marc0der (changes requested):
Can we remove the incorrect entry too?

hgeraldino:
> Can we remove the incorrect entry too?

Ran `uname` on an M1, got `Darwin` and `arm64` as the OS name & Architecture; I amended the PR with those changes.

marc0der (approved):

PR #1171, opened 2023-02-07
Fix main function for native commands

marc0der:
These are steps to improve the main `sdk` function to delegate to the currently available native commands.

- Remove home command hack
- Remove legacy completion command workaround
- Display native help falling back to legacy help
- Give precedence to native home command
- Propagation of up to ten params to a native command

### DIFF
--- a/src/main/bash/sdkman-main.sh
+++ b/src/main/bash/sdkman-main.sh
@@ -141,39 +141,17 @@ function sdk() {
 	# Native commands found under libexec
 	local native_command="${SDKMAN_DIR}/libexec/${COMMAND}"
 	
-	# Internal commands use underscores rather than hyphens
-	local converted_command_name=$(echo "$COMMAND" | tr '-' '_')
-
 	if [ -f "$native_command" ]; then
-		# Available as native command
-		if [ -z "$QUALIFIER" ]; then
-			"$native_command"
-		elif [ -z "$3" ]; then
-			"$native_command" "$QUALIFIER"
-		elif [ -z "$4" ]; then
-			"$native_command" "$QUALIFIER" "$3"
-		elif [ -z "$5" ]; then
-			"$native_command" "$QUALIFIER" "$3" "$4"
-		elif [ -z "$6" ]; then
-			"$native_command" "$QUALIFIER" "$3" "$4" "$5"
-		elif [ -z "$7" ]; then
-			"$native_command" "$QUALIFIER" "$3" "$4" "$5" "$6"
-		elif [ -z "$8" ]; then
-			"$native_command" "$QUALIFIER" "$3" "$4" "$5" "$6" "$7"
-		elif [ -z "$9" ]; then
-			"$native_command" "$QUALIFIER" "$3" "$4" "$5" "$6" "$7" "$8"
-		elif [ -z "$10" ]; then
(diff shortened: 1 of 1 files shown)


hgeraldino (dismissed):
LGTM

src/main/bash/sdkman-main.sh
```diff
@@ -155,8 +152,20 @@ function sdk() {
 			"$native_command" "$QUALIFIER"
 		elif [ -z "$4" ]; then
 			"$native_command" "$QUALIFIER" "$3"
-		else
+		elif [ -z "$5" ]; then
 			"$native_command" "$QUALIFIER" "$3" "$4"
+		elif [ -z "$6" ]; then
+			"$native_command" "$QUALIFIER" "$3" "$4" "$5"
```
helpermethod, line 158: Could you just use a slice, like `"${@:3}"`? Should also work for the other cases.

helpermethod (changes requested):
Looking good so far, but the if-else chain can be improved IMHO.

src/main/bash/sdkman-main.sh
```diff
@@ -155,8 +152,20 @@ function sdk() {
 			"$native_command" "$QUALIFIER"
 		elif [ -z "$4" ]; then
```
helpermethod, line 153: I think you can even reduce the rest of the `elif` chain to

```
elif [[ $# < 10 ]]; then
  "$native_command" "$QUALIFIER" "${@:3}"
else
  ___sdkman_help
fi
```

`${@:3}` is basically a slice from element 3 to the end.

helpermethod (changes requested):
I think there is some room for improvement regarding the elif chain at the end

src/main/bash/sdkman-main.sh
marc0der: Good shout @helpermethod, don't know why I didn't think of that myself :stuck_out_tongue_closed_eyes: 

I've even gone a step further and removed the whole `if` `elif` chain to make it even simpler:

```
"$native_command" "${@:2}"
```

Triggering help at the end isn't a requirement, so I'm just propagating all the parameters now.

src/main/bash/sdkman-main.sh
```diff
@@ -144,27 +141,17 @@ function sdk() {
...
-			"$native_command" "$QUALIFIER" "$3"
-		else
-			"$native_command" "$QUALIFIER" "$3" "$4"
-		fi
-		final_rc=$?
+		"$native_command" "${@:2}"
 
 	elif [ -n "$CMD_FOUND" ]; then
+		# Internal commands use underscores rather than hyphens
+		local converted_command_name=$(echo "$COMMAND" | tr '-' '_')
```
helpermethod, line 149: Better (bc no process is forked)

    converted_command_name=${COMMAND//-/_}

helpermethod (changes requested):
One very minor quibble 😅

src/main/bash/sdkman-main.sh
marc0der: Sure, it was always like that, I only moved it closer to the call site. Feel free to push that change to the feature branch before I merge it.

helpermethod (approved):
LGTM

PR #1180, opened 2023-02-19
Remove broadcast command

marc0der:
This PR removes the `broadcast` command, the `flush broadcast` functionality, and all other broadcast-related code that is now obsolete.


src/main/bash/sdkman-availability.sh
```diff
@@ -16,10 +16,10 @@
 #   limitations under the License.
 #
 
-function __sdkman_update_broadcast_and_service_availability() {
+function __sdkman_update_service_availability() {
 	local broadcast_live_id=$(__sdkman_determine_broadcast_id)
 	__sdkman_set_availability "$broadcast_live_id"
-	__sdkman_update_broadcast "$broadcast_live_id"
+	__sdkman_update_broadcast_id "$broadcast_live_id"
```
marc0der, line 22: This is still needed for the service availability check but will also go away in time. We use the hardcoded nginx `/broadcast/id` mapping here.

src/test/resources/features/checksum_verification.feature
```diff
@@ -61,7 +61,8 @@ Feature: Verify checksums
 		And the candidate "grails" version "1.3.9" is installed
 		And the response headers file is created for candidate "grails" and version "1.3.9"
 		And the exit code is 0
-
+		
+	@manual
```
marc0der, line 65: It seems to be impossible to build this locally. Am I the only one?

I'll be happy to bring this back if this is only occurring on my machine.

/cc @hgeraldino

helpermethod (approved):
LGTM!

src/test/resources/features/checksum_verification.feature
hgeraldino: I'll take a look

hgeraldino (approved):

src/test/resources/features/checksum_verification.feature
marc0der: Thank you!

PR #1187, opened 2023-03-03
Retire auto selfupdate and clean up

marc0der:
- Simplify parameter propagation in the main sdk function.
- Strip out the broadcast id and rename to healthcheck where needed.
- Remove auto-update feature.


helpermethod (approved):
LGTM!

PR #1194, opened 2023-03-18
Selfupdate simplification and version cache removal

marc0der:
- Removes version cache refresh.
- Update sdk env.
- Fix selfupdate.
- Preparation for considering the native version in selfupdate.
- Consider the native version in selfupdate, fix tests and add coverage.
- Clean up.

**Best reviewed commit by commit.**


src/main/bash/sdkman-selfupdate.sh
```diff
@@ -18,17 +18,38 @@
...
+		sdkman_native_version_api="${SDKMAN_CANDIDATES_API}/broker/version/sdkman/native/beta"
+	else
+		sdkman_script_version_api="${SDKMAN_CANDIDATES_API}/broker/version/sdkman/script/stable"
+		sdkman_native_version_api="${SDKMAN_CANDIDATES_API}/broker/version/sdkman/native/stable"
+	fi
+
+	sdkman_remote_script_version=$(__sdkman_secure_curl "$sdkman_script_version_api")
+	sdkman_remote_native_version=$(__sdkman_secure_curl "$sdkman_native_version_api")
+
+	sdkman_local_script_version=$(cat "$SDKMAN_DIR/var/version")
```
helpermethod, line 40: Instead of `$(cat ` you can use `$(< ` here. It's faster and is not affected by alias'ed `cat`.

src/main/bash/sdkman-selfupdate.sh
```diff
@@ -18,17 +18,38 @@
...
+	sdkman_local_native_version=$(cat "$SDKMAN_DIR/var/version_native")
```
helpermethod, line 41: Same here, `$(< ` instead of `$(cat `.

helpermethod (commented):
Some small suggestion, otherwise great work!

helpermethod (approved):
LGTM!

PR #1205, opened 2023-05-13
Remove rosetta2 hack for arm64 on Mac

marc0der:
We introduced this hack when there were very few JDKs available for the ARM64 variant of Mac.

Subsequently, many versions of Java have become available for ARM64 Mac, making this hack redundant. This PR removes all support for this functionality.


eddumelendez (approved):

macalln:
> We introduced this hack when there were very few JDKs available for the ARM64 variant of Mac.
> 
> Subsequently, many versions of Java have become available for ARM64 Mac, making this hack redundant. This PR removes all support for this functionality.

Hi, maybe it’s too early to remove this option? There are still plenty of java native x86 libraries which does not have an arm port. There is also jdk bugs (e.g. Unsafe#getLongVolatile) in arm where it’s nice to be able to switch over to x86.

Issue #1208, opened 2023-06-03
Bug: Install command does not work if SDKMAN_DIR contains whitespace char (user name)


kr5ture:
**Bug report**

I'm using SDKMAN with MSYS2 on Windows. Looks like `sdk install` command fails when current user name contains whitespace characters. System user name is:
```
$ echo $HOME
/home/Janusz Ława
```

Consider following:

* `cd` to `/c/workdir` and `sdk install gradle 7.5.1`:
  ```
  Janusz Ława@DESKTOP-S9IL21H MSYS /c/workdir
  $ sdk install gradle 7.5.1
  ```
* Command fails:
  ```
  $ sdk install gradle 7.5.1
  
  Downloading: gradle 7.5.1
  
  In progress...
  
  ############################################################################################################################################################################## 100.0%
  -bash: /home/Janusz Ława/.sdkman/var/metadata/gradle-7.5.1.headers: No such file or directory
  
  Metadata file not found at '/home/Janusz Ława/.sdkman/var/metadata/gradle-7.5.1.headers', skipping checksum...
  
  Installing: gradle 7.5.1
  Done installing!
  
  
  Setting gradle 7.5.1 as default.
  ln: failed to create symbolic link '/home/Janusz Ława/.sdkman/candidates/gradle/current': No such file or directory
  ```
* It turns out that some unexpected directories were created:
  * Directory named `Ława` within current directory:
    ```
    Janusz Ława@DESKTOP-S9IL21H MSYS /c/workdir
    $ ls -la
    total 8
    drwxr-xr-x 1 Janusz Ława Brak 0 Jun  3 14:45 .
    drwxr-xr-x 1 Janusz Ława Brak 0 Jun  3 14:45 ..
    drwxr-xr-x 1 Janusz Ława Brak 0 Jun  3 14:45 Ława
    ```
  * Directory named `Janusz` within `/home`:
    ```
    Janusz Ława@DESKTOP-S9IL21H MSYS /c/workdir
    $ ls -la /home
    total 16
    drwxr-xr-x 1 Janusz Ława Brak 0 Jun  3 14:51  .
    drwxr-xr-x 1 Janusz Ława Brak 0 Jun  3 14:05  ..
    drwxr-xr-x 1 Janusz Ława Brak 0 Jun  3 14:51  Janusz
    drwxr-xr-x 1 Janusz Ława Brak 0 Jun  3 14:44 'Janusz Ława'
    ```

**To reproduce**

**System info**

[...]

PR #1209, opened 2023-06-03
Add missing quotes when handling file paths within install command

kr5ture:
This PR resolves issues when SDKMAN_DIR has whitespaces (see https://github.com/sdkman/sdkman-cli/issues/1208).


kr5ture:
Wanted to add test case for such scenario but I'm not sure whether I'm able to do so. As far as I can see every test extends from `SdkmanEnvSpecification` which `setup` sections sets `sdkmanBaseDirectory` using static method `prepareBaseDir()` and then rest of the `setup` section relies on it 🤔

marc0der (approved):
LGTM! No need for additional tests as long as existing tests pass. Thanks for contributing!

PR #1245, opened 2023-10-07
fix: Quote `$SDKMAN_DIR` in version subcommand

hyperupcall:
With this change, the `version` subcommand now works if `$SDKMAN_DIR` includes newlines, etc.

### DIFF
--- a/src/main/bash/sdkman-version.sh
+++ b/src/main/bash/sdkman-version.sh
@@ -17,7 +17,7 @@
 #
 
 function __sdk_version() {
-    local version=$(cat $SDKMAN_DIR/var/version)
+	local version=$(cat "$SDKMAN_DIR/var/version")
 	echo ""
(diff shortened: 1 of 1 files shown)


marc0der:
Hi @hyperupcall, the version subcommand has been obsolete for a long time and is replaced by the Rust equivalent over here:
https://github.com/sdkman/sdkman-cli-native/blob/master/src/bin/version/main.rs

hyperupcall:
Thank you for your quick response! I see, the equivalent Rust code is always ran instead (assuming they use the installer and are on one of sdkman-cli-native's 5 supported architectures). In that case, I will close this issue since it doesn't actually change anythig. Maybe there can be a comment at the top or the file can be deleted to prevent confusions like this in the future

marc0der:
No probs, and do let me know if you have any issues. We'll probably begin removing these legacy files soon, but I'll update the project README with some more details.

PR #1269, opened 2024-01-11
Support set -eu

### LINKED ISSUE #597: SDK issues with 'set -eu'
Hi!

At my job I am maintaining several shell scripts created to help software developers to be more productive and I was evaluating the use of SDKMAN to save some work. In order to clean up those scripts, I am using **set -eu** (and other flags that are not relevant at the moment) to be sure that the scripts have all variables it uses and avoid problems with outdated variables being called in some old scripts that might do something unexpected. However, I am facing some difficulties in using **sdk** commands because they rely on unbound variables. For example:

sdkman-sdk-list-fail

Now, let us try installing Kotlin:

sdkman-sdk-install-kotlin-fail

PS: I am working on this fix, as this commit in my fork https://github.com/andretadeu/sdkman-cli/commit/fd26629e2135c55facc2d31f60f7b9ad64809151 shows. However I need to ensure that the tests are correct. If I don't set **set -eu**, all the tests passes, however if I do so, I get some errors and some skipped tests, therefore I am not confident about creating a Pull Request.

PS2: If you have time, could you give an explanation on how these tests works?

Thanks in advance,

André Tadeu de Carvalho

AlexanderYastrebov:
Do not fail when parent script has `set -u` option.

See
* https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html
* https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html
* http://mywiki.wooledge.org/BashFAQ/112

- [x] a GitHub Issue was opened for this feature / bug.

Fixes #597


fredjaya:
Would be good to have this supported!

AlexanderYastrebov:
@marc0der PTAL

PR #1274, opened 2024-01-30
Fix array indexing and length handling

### LINKED ISSUE #1273: Bug: Problems with array indexing and length handling in `sdkman-list.sh`
Cf. https://sdkman.slack.com/archives/CJTNQA94M/p1706573532033689?thread_ts=1706516271.626039&cid=CJTNQA94M

- Array indexing is wrong, it starts at 1 (not 0) and ends at the highest index, not at array length - 1. You have both in C or Java, but shell is different.
- Executed in bash, the `${#versions}` is - for whatever reason - restricted to four bits. So the maximum value is 15. If the versions array is > 15, you have to use `${#versions[@]}` instead (cf. https://linuxhandbook.com/array-length-bash/).

ascheman:
Cf. https://sdkman.slack.com/archives/CJTNQA94M/p1706573532033689?thread_ts=1706516271.626039&cid=CJTNQA94M

- Array indexing is wrong, it starts at 1 (not 0) and ends at the highest index, not at array length -1. You have both in C or Java, but shell is different.
- Executed in bash the `${#versions}` is - for whatever reason - restricted to four bits. So the maximum value is 15. If the versions array is > 15, you have to use `${#versions[@]}` instead (cf. https://linuxhandbook.com/array-length-bash/).

- [x] a GitHub Issue was opened for this feature / bug.

NOTE: I have no idea how to provide a Cucumber test for this case. But I am happy to learn about it, in particular how to test the array size in bash if it is > 15. Feel free to get in touch with me about it.

Fixes #1273

### DIFF
--- a/src/main/bash/sdkman-list.sh
+++ b/src/main/bash/sdkman-list.sh
@@ -76,7 +76,7 @@ function __sdkman_offline_list() {
 	__sdkman_echo_no_colour "--------------------------------------------------------------------------------"
 
 	local versions=($(echo ${versions_csv//,/ }))
-	for ((i = ${#versions} - 1; i >= 0; i--)); do
+	for ((i = ${#versions[@]}; i > 0; i--)); do
 		if [[ -n "${versions[${i}]}" ]]; then
 			if [[ "${versions[${i}]}" == "$CURRENT" ]]; then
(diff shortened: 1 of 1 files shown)


ascheman:
Withdrawn as SDKMAN moves to a different CLI soon

PR #1290, opened 2024-04-16
FIX: Remove download on error

marc0der:


src/main/bash/sdkman-utils.sh
```diff
@@ -32,7 +32,7 @@ function __sdkman_secure_curl() {
 
 function __sdkman_secure_curl_download() {
 	local curl_params
-	curl_params=('--progress-bar' '--location')
+	curl_params=('--progress-bar' '--location' '--remove-on-error')
```
helpermethod, line 35: Hi! I would actually challenge my own suggestion 😄! While that would work for newer versions of `curl`, it may break on older versions, no?

helpermethod: Ah, and it shows 😆 

`curl: option --remove-on-error: is unknown`

helpermethod:
I think a much more portable solution would to perform the cleanup in `sdkman-install.sh`

```
__sdkman_secure_curl_download "${download_url}" --output "${binary_input}" --dump-header "${tmp_headers_file}" || rm -rf "${binary_input}"
```

PR #1325, opened 2024-11-18
use local files instead of downloading from the network

cary929:
In some regions (like China), using curl to download files can be very slow and often interrupted. To solve this issue, you can pre-download the files locally and then use the following method. For example, to install the ant library, the default command is:

```
sdk install ant 1.9.7 [folder]
```

Set the parameter `[folder]` to the string "local" to enable local file installation. The actual command becomes:

```
sdk install ant 1.9.7 local
```

- [ ] a GitHub Issue was opened for this feature / bug.

Fixes #XXX


helpermethod:
Hi @cary929,

this is already possible

https://sdkman.io/usage#install-local-versions

PR #1553, opened 2026-08-08
fix: avoid bash 4-only ${var^^} uppercase expansion on macOS bash 3.2

### LINKED ISSUE #1552: sdkman throws error on macOs when bash is used as default shell
When running bash in macOs ( tested on 26.5, probably older ones also ), https://github.com/sdkman/sdkman-cli/blob/1c8f4cb9101a7cbc2da6453c12bd547531bde29f/src/main/bash/sdkman-path-helpers.sh#L61 fails to work, as this behaviour ( variable^^) only works on bash 4.x, whilst macOs defaults to bash 3.2.57 as of 26.5 

`"-bash: ${candidate_name^^}: bad substitution"`

Using the default behavioir ( when bash or zshrc is not found ), works

fudianchn:
AI disclosure: this change was prepared with AI coding agents, reviewed and revised line by line by me.

## Problem

On macOS the default `/bin/bash` is still **3.2.57**, which does not support the `${var^^}` uppercase parameter expansion (added in bash 4.0). `__sdkman_export_candidate_home` uses it unconditionally when `bash_shell=true`:

https://github.com/sdkman/sdkman-cli/blob/master/src/main/bash/sdkman-path-helpers.sh#L61

```text
-bash: ${candidate_name^^}: bad substitution
```

As a result the `<CANDIDATE>_HOME` variable is never exported on macOS bash, breaking candidate home setup when bash is the default shell (#1552).

## Fix

Keep the `${var^^}` fast path for bash 4+, and on bash < 4 fall back to the same portable `tr '[:lower:]' '[:upper:]'` form already used for the non-bash / non-zsh branch. The bash 4+ path is unchanged.

```diff
-	elif [ "$bash_shell" = true ]; then
-		ucase_candidate_name="${candidate_name^^}"
+	elif [ "$bash_shell" = true ] && [ "${BASH_VERSINFO[0]:-0}" -ge 4 ]; then
+		ucase_candidate_name="${candidate_name^^}"
```

`BASH_VERSINFO[0]` (the major version) is available since bash 2.0, so the guard is safe on the 3.2 we are protecting.

## Verification

The CI Cucumber suite runs bash on Linux (bash 4+), where `${var^^}` works, so it cannot exercise the bash 3.2 path. I verified under **bash 3.2.57 in `docker bash:3.2`** (the exact macOS default):

```text
bash_major=3  full=3.2.57(1)-release
OLD ^^     -> /t.sh: line 6: ${candidate_name^^}: bad substitution
NEW fixed  -> result='JAVA'
```

And confirmed no regression on bash 5.2 (host), where both the old and new paths yield `JAVA` (the `^^` fast path is retained).

`bash -n` passes on the modified script.

Closes #1552


bertjan:
I ran into the same issue (OSX 26.6.2, bash 3.2.57).
Tried the fix in https://github.com/sdkman/sdkman-cli/pull/1553/commits/121e999d34d8c89ffea3bb6f6652bd9021b1e5f5 by manually making the change in `sdkman-path-helpers.sh`  - can confirm the fix works for me. Thanks!

PR #1556, opened 2026-08-24
Revert "fix: install tarballs served through the renaming post-install hook"

marc0der:
Reverts sdkman/sdkman-cli#1555


asm0dey:
Was it a wrong way?

marc0der:
Hey, sorry about that. Chloé merged that one by accident. It has already been fixed in sdkman-hooks over here:
https://github.com/sdkman/sdkman-hooks/pull/96

I really appreciate you looking into this, though. Thanks for taking the time to do it!

asm0dey:
We were affected, so I didn't really have a choice! Thanks for fixing it! When can I expect it to be GA?

PR #1562, opened 2026-09-10
Fix #1561 by making the PATH entry replacement regex more expressive

### LINKED ISSUE #1561: Bug: Merged PATH entries after running `sdk use`
**Bug report**

After running `sdk use` on a candidate not having a **bin** directory, multiple **PATH** entries will get merged together.

**To reproduce**

This issue is not trivial to detect, as most SDKs actually have a **bin** directory - which is why I discovered it by chance.

However, just to see it in practice, please consider these steps:

1. Install SDKMAN from scratch

2. Run `sdk install java 23-open`

3. Run `rm -r ~/.sdkman/candidates/java/23-open/bin`

4. Open a new Bash shell

5. Run `sdk install java 25.0.2-open`

6. Run `echo $PATH | tr ':' '\n'`  - the PATH entries should still be fine

6. Run `sdk use java 23-open`

7.  Run `echo $PATH | tr ':' '\n'` again: the PATH entry related to **java/23-open** is merged with the following one

**System info**

* **OS**: Linux (all versions)
* **Shell**: Bash (all versions)
* **SDKMAN version**: 5.23.0

giancosta86:
Fixes #1561.

The original version of the regex expects SDKMAN-generated entries to **always** include a **/bin** subdirectory - which might well be the most common case (e.g., Java, Maven, Gradle)... however, the **__sdkman_prepend_candidate_to_path** function explicitly supports candidates without such **bin**.

Consequently, by adding the `:` delimiter - used to separate **PATH** entries - to the exclusion group, the revised regex now correctly supports both scenarios.

The result can be tested by trying to reproduce the bug; an automated test might well be interesting, although that would probably require introducing dedicated Cucumber verbs - or, at least, finding an SDK not having the **bin** directory.


marc0der:
@claude please review this PR

marc0der:
@giancosta86 This is now deployed on our beta environment. Would you be willing to try it to see if it resolves the issue? Once you confirm, I will release that to stable.

giancosta86:
@marc0der 🤠It works! Thanks a lot! 🥳

src/main/bash/sdkman-use.sh
```diff
@@ -37,7 +37,7 @@ function __sdk_use() {
 	# Just update the *_HOME and PATH for this shell.
 	__sdkman_set_candidate_home "$candidate" "$version"
 
-	if [[ $PATH =~ ${SDKMAN_CANDIDATES_DIR}/${candidate}/([^/]+) ]]; then
+	if [[ $PATH =~ ${SDKMAN_CANDIDATES_DIR}/${candidate}/([^/:]+) ]]; then
```
MasterDisaster, line 40: expression ending in "([^/:]+)" -> allows also whitespace characters (and/or terminal control characters), this is not the best way to Decide it has proper finish and not finishing PATH element with forbidden "/" right before ":". This is not gone work properly on OLD terminals or customized "echo" builtin environments.

giancosta86: 🫖I'm afraid that:

1. the claim should be properly proven, maybe suggesting an alternative approach – but the project is now in maintenance mode;

2. the point does not affect the current patch in itself, which merely fixes an existing issue in a consistent and reproducible way.

marc0der:
@giancosta86 thanks for contributing! I'll promote this to our stable environment today.

src/main/bash/sdkman-use.sh
MasterDisaster: yes - agree.
I always ten do write code that is "stupidity immune". One of users mentioned "no bin" folder in SDKs so this made the "/" right before ":" of PATH values list. This was not my case but not ONCE I have seen BOM signs start to appear in STRING values:
#1. person A edited sth
#2. person B took a peek and used it's value
#3. person C sees "unwanted space that is a transcribed BOM signs unseen before

You got here a nice solution, that has a opinion of stable. Hiding behind (not in support) makes it less predictably stable in next spinn.
People that do not handle well complexity of packages in 90% reuse Your tool and those people tent to make VERY strange mistakes during it's work (no knowledge).
They work in strange terminals on windows (CRLF) and have strange terminals on "almost linux" os and usually are Copy + Paste 'ers with work on paths corrupting ASCI strings.
The PATH variable sanitization is in my opinion a CRITICAL topic. 
Other Paths on OS (@config) or derived from env environments are NOT that super important as PATH var is. 
a BIT of more caution is advised and to suite with more security in future and have less problems in solutions spinning a new base from Your tool. 
Thank You for reading this.

